PR #7, opened 2018-09-13
added feature: if there's a config.xml add projects to database

mplieske:


src/test/java/configTest.xml
```diff
@@ -0,0 +1,37 @@
...
+  
+    <!-- One default project should exist (or it is created). It is used as active project when starting the application -->
+    <project isDefault="true">
+      <name>Idle</name>
+      <isWork>false</isWork>
+      <color>BROWN</color>
+    </project>
+  
+    <project>
+      <name>Carl Zeiss</name>
```
Death111: please use some other project name

src/main/java/de/doubleslash/keeptime/common/ConfigParser.java
```diff
@@ -0,0 +1,76 @@
...
+         for (int temp = 0; temp < nList.getLength(); temp++) {
+            final Node nNode = nList.item(temp);
+            if (nNode.getNodeType() == Node.ELEMENT_NODE) {
+               final Element eElement = (Element) nNode;
+               // get element
+               final String name = eElement.getElementsByTagName("name").item(0).getTextContent();
+               final String isWork = eElement.getElementsByTagName("isWork").item(0).getTextContent();
+               final String color = eElement.getElementsByTagName("color").item(0).getTextContent();
+
+               System.out.println("checking for: " + name);
```
Death111: please use a logger instead for sysout

src/main/java/de/doubleslash/keeptime/common/ConfigParser.java
```diff
@@ -0,0 +1,76 @@
...
+
+import de.doubleslash.keeptime.controller.Controller;
+import de.doubleslash.keeptime.model.Model;
+import de.doubleslash.keeptime.model.Project;
+import javafx.scene.paint.Color;
+
+public class ConfigParser {
+
+   Controller controller;
+   Model model;
```
Death111: you don't use the model in the class - delete it

src/test/java/de/doubleslash/keeptime/common/ConfigParserTest.java
```diff
@@ -0,0 +1,192 @@
...
+   private final String CONFIG_FILENAME = classLoader.getResource("configTest.xml").getFile();
+
+   private final Project idle = new Project("Idle", Color.AQUAMARINE, true, 0);
+   private final Project heinz = new Project("Heinz", Color.BLUEVIOLET, true, 1);
+   private final Project peter = new Project("Peter", Color.CRIMSON, true, 2);
+   private final Project hinzUndKunz = new Project("Hinz und Kunz", Color.DARKRED, true, 3);
+   private final ObservableList<Project> projects = FXCollections.observableArrayList();
+
+   Model model = new Model();
+   private final ObservableList<Project> spyModel = spy(model.availableProjects);
```
Death111: nice :) But as we mock the Controller, we don't have to test if the controller writes the new projects into the model correctly. We should just make sure, that controller.addNewProject is called as often as we expect with the needed parameters So we could get rid of all references of the model.

src/main/java/de/doubleslash/keeptime/common/ConfigParser.java
```diff
@@ -0,0 +1,76 @@
...
+      this.model = model;
+      this.controller = controller;
+   }
+
+   public static boolean hasConfigFile(final String fileName) {
+      final File f = new File(fileName);
+      return f.exists();
+   }
+
+   public void parserConfig(final File inputFile) {
```
Death111: could be renamed to parseConfig

src/test/java/de/doubleslash/keeptime/common/ConfigParserTest.java
```diff
@@ -0,0 +1,192 @@
...
+            Log.debug("called with arguments: " + Arrays.toString(args));
+            projects.add(new Project((String) args[0], (Color) args[2], (Boolean) args[1], (Integer) args[3]));
+            return "";
+
+         }
+      }).when(controller).addNewProject(anyString(), anyBoolean(), any(Color.class), anyInt());
+
+      loadConfigFile(CONFIG_FILENAME);
+
+      verify(controller).addNewProject(eq("Carl Zeiss"), anyBoolean(), any(Color.class), anyInt());
```
Death111: there you should verify if the isWork, color, index are the ones you expect the loader to create (like with the name)

src/test/java/de/doubleslash/keeptime/common/ConfigParserTest.java
```diff
@@ -0,0 +1,192 @@
...
+            Log.debug("called with arguments: " + Arrays.toString(args));
+            projects.add(new Project((String) args[0], (Color) args[2], (Boolean) args[1], (Integer) args[3]));
+            return "";
+
+         }
+      }).when(controller).addNewProject(anyString(), anyBoolean(), any(Color.class), anyInt());
+
+      loadConfigFile(CONFIG_FILENAME);
+
+      verify(controller, atLeastOnce()).addNewProject(eq("Peter Lustig"), anyBoolean(), any(Color.class), anyInt());
```
Death111: I wonder if this would return true, if the addNewProjects was not called in this test, as it was called in another test before. As you use the same controller instance for all tests? I think you should create a new mock-controller for each test case e.g. with @Before

src/test/java/de/doubleslash/keeptime/common/ConfigParserTest.java
```diff
@@ -0,0 +1,192 @@
...
+   private final Project heinz = new Project("Heinz", Color.BLUEVIOLET, true, 1);
+   private final Project peter = new Project("Peter", Color.CRIMSON, true, 2);
+   private final Project hinzUndKunz = new Project("Hinz und Kunz", Color.DARKRED, true, 3);
+   private final ObservableList<Project> projects = FXCollections.observableArrayList();
+
+   Model model = new Model();
+   private final ObservableList<Project> spyModel = spy(model.availableProjects);
+   private final Controller controller = mock(Controller.class);
+
+   @Before
```
Death111: before is run before each single test. so after n tests there should be n*4 projects in the projects list. I guess that is not what you wanted :) beforeclass can be used in this case, where the method is called just once before all tests

Death111:
Also there should be some mechanism to ask the user if he wants to import the config, but not each time he starts the app :)

src/test/java/de/doubleslash/keeptime/common/ConfigParserTest.java
mplieske: no, junit '@Test' annotation  instatiates a new Object of your test class for each test :)

PR #23, opened 2019-04-03
Feature/settings color timeline

mplieske:
added a colorized timeline to the report and put logic for in in separate class


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -151,6 +164,10 @@ private void updateReport(final LocalDate newvalue) {
 
       this.currentDayTimeLabel.setText(DateFormatter.secondsToHHMMSS(currentSeconds));
       this.currentDayWorkTimeLabel.setText(DateFormatter.secondsToHHMMSS(currentWorkSeconds));
+
+      controller = new Controller(model, dateProvider);
```
Death111: we should not create new instances of the controller. if the controller is needed we should give a reference to it like for the model

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -151,6 +164,10 @@ private void updateReport(final LocalDate newvalue) {
...
+      colorTimeLine = new ColorTimeLine(reportColorTimeLine, model, controller);
+      colorTimeLine.update();
```
Death111: what happens if I change the day? will the timeline update to this day?

src/main/java/de/doubleslash/keeptime/view/ViewController.java
```diff
@@ -316,7 +316,8 @@ private void initialize() {
             label.setText(DateFormatter.secondsToHHMMSS(seconds));
          }
 
-         updateProjectColorTimeline();
+         final ColorTimeLine mainColorTimeLine = new ColorTimeLine(canvas, model, controller);
```
Death111: why recreate the timeline each time? just call update on an existing instance :)

src/main/java/de/doubleslash/keeptime/common/ColorTimeLine.java
```diff
@@ -0,0 +1,42 @@
...
+import javafx.scene.paint.Color;
+
+public class ColorTimeLine {
+
+   private final Canvas canvas;
+   private final Model model;
+   private final Controller controller;
+
+   public ColorTimeLine(final Canvas canvas, final Model model, final Controller controller) {
+      this.canvas = canvas;
```
Death111: we should not have a dependency to the model or the controller. we should be able to tell it the needed things to draw. the class should not have to calculate this on its own.

src/main/java/de/doubleslash/keeptime/common/ColorTimeLine.java
```diff
@@ -0,0 +1,42 @@
+package de.doubleslash.keeptime.common;
```
Death111: this should be, like the LicenseTableRow class just in the UI package - as this is just custom UI related

src/main/java/de/doubleslash/keeptime/common/ColorTimeLine.java
mplieske: DONE! noice :)

src/main/java/de/doubleslash/keeptime/view/ReportController.java
mplieske: It didn't but now it does 😃

mplieske:
im ready :)

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -237,29 +236,35 @@ public void setComment(final String notes) {
...
+      final List<Work> workItems = new ArrayList<>();
 
-      return model.getPastWorkItems().stream().filter(work -> {
-         final Project project = work.getProject();
-         // find up to date reference to project
-         final Optional<Project> optionalProject = model.getAllProjects().stream()
-               .filter(p -> p.getId() == project.getId()).findAny();
-         if (optionalProject.isPresent()) {
-            return optionalProject.get().isWork();
+      for (final Work w : model.getPastWorkItems()) {
```
Death111: please reuse the old logic with the id equals. Not sure why i did it like this, but it was working :) (u can still use the for each)

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -92,7 +113,10 @@ private void updateReport(final LocalDate newvalue) {
       for (final Project project : workedProjectsSet) {
          final Label projectName = new Label(project.getName());
          projectName.setFont(labelFontBold);
-         this.gridPane.add(projectName, 0, rowIndex);
+         final Circle circle = new Circle(5, project.getColor());
+
+         this.gridPane.add(circle, 0, rowIndex);
+         this.gridPane.add(projectName, 1, rowIndex);
```
Death111: please use a HBox for this, instead of adding a new column. so the color circle and project name are in the same column :)

src/main/java/de/doubleslash/keeptime/controller/Controller.java
mplieske: Ok I understand why :)

src/main/java/de/doubleslash/keeptime/view/ReportController.java
mplieske: good idea, props 👍

mplieske:
Merging would be nice at this point 😁

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -253,29 +252,38 @@ public void setComment(final String notes) {
...
-      return model.getPastWorkItems().stream().filter(work -> {
+      for (final Work work : model.getPastWorkItems()) {
          final Project project = work.getProject();
-         // find up to date reference to project
-         final Optional<Project> optionalProject = model.getAllProjects().stream()
-               .filter(p -> p.getId() == project.getId()).findAny();
-         if (optionalProject.isPresent()) {
-            return optionalProject.get().isWork();
+         for (final Project p : model.getAllProjects()) {
+            if (p.getId() == project.getId() && project.isWork()) {
```
Death111: this should be p.isWork, otherwise this does not help

src/main/resources/layouts/report.fxml
```diff
@@ -100,9 +106,9 @@
                <content>
                   <GridPane fx:id="gridPane" prefWidth="498.0">
                     <columnConstraints>
-                      <ColumnConstraints hgrow="SOMETIMES" maxWidth="349.0" minWidth="10.0" prefWidth="290.0" />
-                      <ColumnConstraints hgrow="SOMETIMES" maxWidth="243.0" minWidth="10.0" prefWidth="131.0" />
-                        <ColumnConstraints hgrow="SOMETIMES" maxWidth="243.0" minWidth="10.0" prefWidth="77.0" />
+                      <ColumnConstraints hgrow="SOMETIMES" maxWidth="374.0" minWidth="0.0" prefWidth="216.0" />
+                      <ColumnConstraints hgrow="SOMETIMES" maxWidth="400.0" minWidth="0.0" prefWidth="146.0" />
+                        <ColumnConstraints hgrow="SOMETIMES" maxWidth="258.0" minWidth="0.0" prefWidth="102.0" />
```
Death111: brauchen wir die changes hier? da wir ja keine neue spalte haben, sollten die alten werte ok sein oder?

mplieske: nope weil wir ja jetzt eine spalte weniger haben

PR #34, opened 2019-09-02
added shortcut (win + down) to minimize main window

mplieske:


src/main/java/de/doubleslash/keeptime/Main.java
```diff
@@ -260,6 +266,31 @@ public void handle(final WindowEvent event) {
 
    }
 
+   private void registerMinifyEventlistener(final Scene mainScene, final Stage primaryStage) {
+      final KeyCode down = KeyCode.DOWN;
+      final KeyCombination keyComb = new KeyCodeCombination(down, KeyCombination.META_DOWN);
+      mainScene.addEventFilter(KeyEvent.KEY_RELEASED, keyEvent -> {
+         LOG.info("Key was Pressed '{}'", keyEvent.getCode());
```
Death111: now we actually have a keylogger :)

src/main/java/de/doubleslash/keeptime/Main.java
```diff
@@ -260,6 +266,31 @@ public void handle(final WindowEvent event) {
 
    }
 
+   private void registerMinifyEventlistener(final Scene mainScene, final Stage primaryStage) {
+      final KeyCode down = KeyCode.DOWN;
+      final KeyCombination keyComb = new KeyCodeCombination(down, KeyCombination.META_DOWN);
```
Death111: cool, didn't know this 👍  should also use this in the listener for the popup

mplieske: instead of jnative hook?

PR #36, opened 2019-09-26
Feature/searchbar for projects in main view

mplieske:
- styling of popup is consistent with main style
 - styling of scrollbar is consistent with main style
 - popup has same functionality as main project list
 - main project list has searchbar now


src/main/java/de/doubleslash/keeptime/view/ViewController.java
```diff
@@ -97,15 +87,33 @@
...
+      double y;
+   }
+
+   private static final Logger LOG = LoggerFactory.getLogger(ViewController.class);
+
    private static final String TIME_ZERO = "00:00:00";
 
+   private final Delta dragDelta = new Delta();
+
+   private final Canvas taskbarCanvas = new Canvas(32, 32);
```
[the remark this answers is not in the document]
mplieske: Noted 😃

mplieske:
- added tooltip so that the project name is displayed even if to long for the project list

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -142,6 +142,11 @@ public void deleteProject(final Project p) {
          LOG.error("You cannot delete the default project. Tried to delete project '{}'", p);
          return;
       }
+
+      if (p == model.activeWorkItem.get().getProject()) {
```
Death111: not clear what this does. add a comment, but better a function like isProjectActive(p)

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -0,0 +1,390 @@
...
+            super.updateItem(item, empty);
+
+            setText("");
+            if (item == null || empty) {
+               setGraphic(null);
+            } else {
+               LOG.debug("Adding tooltip to Project Label.");
+               final Pane pane = (Pane) projectSelectionNodeMap.get(item);
+               final Label projectNameLabel = (Label) pane.getChildren().get(0);
+               projectNameLabel.setTooltip(new Tooltip(item.getName()));
```
Death111: why set a tooltip everytime here and not while creating the projectSelectionNodeMap item

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -0,0 +1,390 @@
...
+               LOG.trace("Item: '{}' -> '{}'", item.getName(), projectSelectionNodeMap.get(item));
+               setGraphic(projectSelectionNodeMap.get(item));
```
Death111: if a project is added, a NPE is thrown, as this project does not have a node map

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -69,18 +70,23 @@ public ProjectsListViewController(final Model model, final Controller controller
...
-         if (project.isEnabled()) {
-            final Node node = addProjectToProjectList(project);
-            projectSelectionNodeMap.put(project, node);
+      model.getSortedAvailableProjects().addListener((ListChangeListener<? super Project>) listener -> {
+         listener.next();
+         if (listener.wasAdded()) {
+            final Project addedProject = listener.getAddedSubList().get(0);
+            addProjectToProjectSelectionNodeMap(addedProject);
          }
+      });
```
mplieske: Put the logic into the updateItem method. If it finds an entry in the nodemap use it, otherwise create it. 
Like it is now, it is quite complex and dependent of each other (this listener has to be exectued before the updateItem - which we cannot guarantee)

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -151,13 +157,26 @@ public ProjectsListViewController(final Model model, final Controller controller
       searchTextField.setPromptText("Search");
    }
 
+   private void addProjectToProjectSelectionNodeMap(final Project project) {
+      if (project.isEnabled()) {
+         final Pane projectElement = addProjectToProjectList(project);
+         final Label projectNameLabel = (Label) projectElement.getChildren().get(0);
+         projectNameLabel.setTooltip(new Tooltip(projectNameLabel.getText()));
```
mplieske: I meant to put the tooltip logic into the "addProjectToProjectList" method, as this creates the ui representation of the project (which should also include the tooltip).
When doing this, please rename the method also, as it currently does not do what it says.

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -0,0 +1,395 @@
...
+            if (item == null || empty) {
+               setGraphic(null);
+            } else {
+               final Pane graphic = (Pane) projectSelectionNodeMap.get(item);
+               if (graphic == null) {
+                  final Project addedProject = item;
+                  addProjectToProjectSelectionNodeMap(addedProject);
+               }
+               LOG.trace("Item: '{}' -> '{}'", item.getName(), projectSelectionNodeMap.get(item));
+               setGraphic(projectSelectionNodeMap.get(item));
```
Death111: please use the same variable when logging. If you change one, the other will not change with it

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -0,0 +1,395 @@
...
+            searchTextField.setText("");
+         }
+      });
+
+      availableProjectsListView.getSelectionModel().setSelectionMode(SelectionMode.SINGLE);
+      availableProjectsListView.getSelectionModel().selectFirst();
+   }
+
+   private void addProjectToProjectSelectionNodeMap(final Project project) {
+      if (project.isEnabled()) {
```
Death111: remove this condition, as this may result in unexpected behaviour (if you add something and it is not added). The method also does something, it does not tell you (only adding if its enabled) :)

PR #42, opened 2020-03-09
Feature/#28 edit work times

JSchraff:


src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -0,0 +1,260 @@
...
+      startDatePicker.setValue(work.getStartTime().toLocalDate());
+      endDatePicker.setValue(work.getEndTime().toLocalDate());
+
+      startTimeSpinner.getEditor().textProperty().addListener((observable, oldValue, newValue) -> {
+         final LocalTimeStringConverter stringConverter = new LocalTimeStringConverter(FormatStyle.MEDIUM);
+         final StringProperty text = (StringProperty) observable;
+         try {
+            stringConverter.fromString(newValue);
+            text.setValue(newValue);
+            startTimeSpinner.increment(0); // TODO find better Solution
```
Death111: please also add why we need this line

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -184,17 +184,47 @@ public void editProject(final Project projectToBeUpdated, final Project newValue
       model.getProjectRepository().saveAll(changedProjects);
    }
 
+   public void editWork(final Work workToBeEdited, final Work newValuedWork) {
+      LOG.info("Changing work '{}' to '{}'.", workToBeEdited, newValuedWork);
+      int index = model.getPastWorkItems().size() - 1;
+      for (int i = 0; i < model.getPastWorkItems().size(); i++) {
```
Death111: Resorting the pastWorkitems manualy is extra code we do not need. I think its just for the timeline to be displayed correctly?

Better use a sorted list (see e.g. https://github.com/doubleSlashde/KeepTime/blob/develop/src/main/java/de/doubleslash/keeptime/model/Model.java#L61 ) or sort if if we need it sorted.

src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -0,0 +1,260 @@
...
+            return new ListCell<Project>() {
+
+               @Override
+               protected void updateItem(final Project item, final boolean empty) {
+                  super.updateItem(item, empty);
+                  if (item == null || empty) {
+                     setGraphic(null);
+                  } else {
+                     final Color color = item.getColor();
+                     final String hexColor = String.format("#%02X%02X%02X", (int) (color.getRed() * 255),
```
Death111: FYI: Theres a class ColorHelper. Please extract the code here and create a function like colorToHex :)
https://github.com/doubleSlashde/KeepTime/blob/develop/src/main/java/de/doubleslash/keeptime/common/ColorHelper.java

BUT not needed as the logic already exists:
Better use the StyleUtils to change the bg color like in:
https://github.com/doubleSlashde/KeepTime/blob/develop/src/main/java/de/doubleslash/keeptime/view/ViewController.java#L339 
:)

src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -0,0 +1,260 @@
...
+
+         @Override
+         public Project fromString(final String string) {
+            return null;
+         }
+      });
+
+      projectComboBox.getSelectionModel().select(work.getProject());
+      final Color color = work.getProject().getColor();
+      final String hexColor = String.format("#%02X%02X%02X", (int) (color.getRed() * 255),
```
Death111: duplicated code 3 times. In general we should eliminate duplicate code by extracting into a method :) But see comment above.

src/main/java/de/doubleslash/keeptime/controller/Controller.java
JSchraff: sorted list is impossible because it is an Observable list.

if we resort it everytime we need it sorted would sort every second to update timeline

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -186,32 +186,22 @@ public void editProject(final Project projectToBeUpdated, final Project newValue
...
-            model.getPastWorkItems().remove(work);
-            continue;
-         }
-         if (work.getStartTime().isAfter(newValuedWork.getStartTime())) {
-            index = i;
-            break;
-         }
-      }
+
+      model.getPastWorkItems().removeIf(w -> (w.getId() == workToBeEdited.getId()));
```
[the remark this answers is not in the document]
Death111: Side note: please don't resolve the conversations on your own - the person who created it should also resolve it.

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -197,6 +234,71 @@ private void updateReport(final LocalDate dateToShow) {
...
+      final ManageWorkController manageWorkController = loader.getController();
+      manageWorkController.setModel(model);
+      manageWorkController.initializeWith(work);
+
+      dialog.setResultConverter(dialogButton -> {
+         if (dialogButton == ButtonType.OK) {
+            return manageWorkController.getWorkFromUserInput();
+         }
+         // TODO: Do you really want to return null?
+         return null;
```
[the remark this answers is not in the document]
JSchraff: i think that makes everything to convoluted as we get an Optional<Optional> result from the dialog and would need to test double if is present

src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -0,0 +1,276 @@
...
+            // ignore selectionChanges
+            if (comboChange) {
+               comboChange = false;
+               isValidProject = true;
+
+               return;
+            }
+
+            // is necessary to not autoselect same Project if Project was selected
+            if (Boolean.TRUE.equals(isValidProject)) {
```
Death111: why this complicated condition? if(isValidProject) ?

src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -0,0 +1,276 @@
...
+               }
+
+            });
+
+         }
+      });
+
+      // manages Focusbehaviour
+      projectComboBox.getEditor().focusedProperty().addListener((final ObservableValue<? extends Boolean> observable,
+            final Boolean oldIsFocused, final Boolean newIsFocused) -> {
```
Death111: fyi: you should also be able to omit the types like (observable, oldIsFocused, newIsFocused) ->

PR #45, opened 2020-03-10
Feature/optional note reminder

JSchraff:


src/main/java/de/doubleslash/keeptime/model/Model.java
```diff
@@ -76,6 +77,9 @@ public Model(final ProjectRepository projectRepository, final WorkRepository wor
    public final ObjectProperty<Boolean> useHotkey = new SimpleObjectProperty<>(false);
    public final ObjectProperty<Boolean> displayProjectsRight = new SimpleObjectProperty<>(false);
    public final ObjectProperty<Boolean> hideProjectsOnMouseExit = new SimpleObjectProperty<>(true);
+   public final ObjectProperty<Boolean> emptyNoteReminder = new SimpleObjectProperty<>(false);
```
Death111: Name of the variable is not nice. As it is a boolean value it should 'answer a question' lik e.g. hintIfNoteIsEmpty -> yes or no :)

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -165,7 +166,25 @@ private void changeProject(final Project newProject, final long minusSeconds) {
       if (hideable) {
          mainStage.hide();
       }
+      if (model.emptyNoteReminder.get()) {
+         final Work currentWork = model.activeWorkItem.get();
+         if (currentWork != null && currentWork.getNotes().isEmpty()) {
+            final Alert alert = new Alert(AlertType.CONFIRMATION);
```
Death111: Please adapt the dialog so the user can directly add a note. More userfriendly. This was a 'feature' in a eariler version which was removed as there was no setting yet. See: https://github.com/doubleSlashde/KeepTime/blob/master/src/main/java/de/doubleslash/keeptime/viewpopup/ViewControllerPopup.java#L65
Input field should be focused so the user can directly type and press enter to confirm the switch (with our without added note :))

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -332,14 +352,13 @@ private String createTooltipTextForProject(final Project p) {
    private GridPane setUpEditProjectGridPane(final Project p, final Dialog<Project> dialog) {
       GridPane grid;
       final FXMLLoader loader = new FXMLLoader(Resources.getResource(RESOURCE.FXML_MANAGE_PROJECT));
+      loader.setControllerFactory(model.getSpringContext()::getBean);
```
Death111: we should extract the duplicate loader functionality some time into another class

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -164,7 +176,7 @@ private void initialize() {
          controller.updateSettings(hoverBackgroundColor.getValue(), hoverFontColor.getValue(),
```
Death111: we should pass an object as param instead of all values on each own. Like for editWork passing a Settings object with the updated Settings as param, instead of each param of settings on each own

src/main/resources/db/migration/V1_1_1__added_note_reminder_setting.sql
```diff
@@ -0,0 +1,2 @@
+ ALTER TABLE settings
+  ADD COLUMN empty_note_reminder BOOLEAN NOT NULL DEFAULT(false)
```
Death111: cool

src/main/java/de/doubleslash/keeptime/Main.java
```diff
@@ -89,7 +87,7 @@ public void init() throws Exception {
       springContext = SpringApplication.run(Main.class);
       // TODO test if everywhere is used the same model
       model = springContext.getBean(Model.class);
-      controller = springContext.getBean(Controller.class);
+      model.setSpringContext(springContext);
```
Death111: you included the dependency injection refactoring into this feature. should avoid mixing changes as it makes review harder and if one aspect is not ok, both features (here DI, Note reminder) are blocked from being merged.
Don't need to change it - just for later :)

src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
Death111: Please add the current project name to the dialog :) Similar to how it is in the linked code. May forget what project is currently active :D

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
JSchraff: changed in #49

Death111:
waiting for #47

Death111:
text control for the notes is not a textarea like in the other controls (main window, edit work dialog), but ok for me

PR #46, opened 2020-03-10
Feature/rework work items in report ui

JSchraff:


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -87,130 +96,208 @@
...
-      });
+   private void initTableView() {
+      final TreeTableColumn<TableRow, String> noteColumn = new TreeTableColumn<>("Notes");
+      noteColumn.setCellValueFactory(new TreeItemPropertyValueFactory<TableRow, String>("notes"));
+      noteColumn.setMinWidth(200);
+      this.workTableTreeView.getColumns().add(noteColumn);
+
+      final TreeTableColumn<TableRow, String> timeRangeColumn = new TreeTableColumn<>("Timeslot");
+      timeRangeColumn.setCellValueFactory(new TreeItemPropertyValueFactory<TableRow, String>("timeRange"));
+      timeRangeColumn.setMinWidth(100);
```
Death111: it is always cutting the timeslot strings. please make it wider. We may also make the stage a little wider if need be

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -87,130 +96,208 @@
...
+      colorTimeLine = new ColorTimeLine(colorTimeLineCanvas);
+   }
 
-      this.datePicker = new DatePicker(LocalDate.now());
-      this.datePicker.valueProperty().addListener((observable, oldvalue, newvalue) -> {
-         LOG.info("Datepicker selected value changed to {}", newvalue);
-         updateReport(newvalue);
-      });
+   private void initTableView() {
+      final TreeTableColumn<TableRow, String> noteColumn = new TreeTableColumn<>("Notes");
```
Death111: the notes are not wrapped anymore, but shown with ...
should wrap the notes again if possible, so that they can be read
-was commentLabel.setWrapText(true);

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -87,130 +96,208 @@
...
+      noteColumn.setCellValueFactory(new TreeItemPropertyValueFactory<TableRow, String>("notes"));
+      noteColumn.setMinWidth(200);
+      this.workTableTreeView.getColumns().add(noteColumn);
```
Death111: Projects which are work should use a underline like before:
 projectName.setUnderline(project.isWork());

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -87,130 +96,208 @@
...
+      timeRangeColumn.setCellValueFactory(new TreeItemPropertyValueFactory<TableRow, String>("timeRange"));
+      timeRangeColumn.setMinWidth(100);
+      this.workTableTreeView.getColumns().add(timeRangeColumn);
+
+      final TreeTableColumn<TableRow, String> timeSumColumn = new TreeTableColumn<>("Duration");
+      timeSumColumn.setCellValueFactory(new TreeItemPropertyValueFactory<TableRow, String>("timeSum"));
+      timeSumColumn.setMinWidth(100);
+      this.workTableTreeView.getColumns().add(timeSumColumn);
+
+      final TreeTableColumn<TableRow, Button> buttonColumn = new TreeTableColumn<>();
```
Death111: maybe use 'Controlls' as heading?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
Death111: --  Duration is always to big :) so we can make duration smaller

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -113,120 +115,118 @@
...
-
-            contextMenu.getItems().add(editMenuItem);
-
-            clickDummy.setOnContextMenuRequested(
-                  event -> contextMenu.show(clickDummy, event.getScreenX(), event.getScreenY()));
+         for (final Work w : onlyCurrentProjectWork) {
+            final HBox workButtonBox = new HBox();
+            workButtonBox.getChildren().add(createEditWorkButton(w));
+            final TreeItem<TableRow> workRow = new TreeItem<>(new WorkTableRow(w, workButtonBox));
+            projectRow.getChildren().add(workRow);
```
Death111: the rows are now to high. can only fit < half rows than before. please decrease the height of the rows ( i guess the button is the problem)

PR #47, opened 2020-03-12
Feature/dependency injection

JSchraff:


Death111:
looks good so far. Merge conflicts have to be resolved and I'll have a look again.

src/main/java/de/doubleslash/keeptime/viewpopup/ViewControllerPopup.java
```diff
@@ -115,6 +107,15 @@ private void initialize() {
 
       Interval.registerCallBack(() -> projectsListViewController.tick());
 
+      projectsListViewController = new ProjectsListViewController(model, controller, stage, projectListView,
```
Death111: stage is null here, resulting in npe with popupstage when "change with time" is used.
we may still need the 2nd initialize here?

src/main/java/de/doubleslash/keeptime/viewpopup/ViewControllerPopup.java
```diff
@@ -115,6 +107,15 @@ private void initialize() {
 
       Interval.registerCallBack(() -> projectsListViewController.tick());
```
Death111: if we look at it like this, this seems dangerous, registering callbacks on an object, which is instantiated one line after the registration :)

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -101,22 +104,31 @@
...
+
    @FXML
    private void initialize() {
       LOG.info("Init reportController");
       currentReportDate = LocalDate.now();
-      initTableView();
       colorTimeLine = new ColorTimeLine(colorTimeLineCanvas);
+
+      loadCalenderWidget();
+      initTableView();
```
Death111: can you please have a look if I merged this correctly? It is working :)

src/main/java/de/doubleslash/keeptime/viewpopup/ViewControllerPopup.java
JSchraff: yes it is merged correctly

PR #48, opened 2020-03-13
added spring repo to resolve mvn resolve error

Death111:
when building with mvn e.g. 'mvn install' the fontawesome libs cannot be resolved. 
The referenced libraries are not in the central repo, but in spring repo: https://mvnrepository.com/artifact/de.jensd/fontawesomefx-fontawesome?repo=springio-plugins-release
the versions in central are not compatible with java 8. seems like we have to include the spring repo

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,12 +31,12 @@
     <relativePath /> <!-- lookup parent from repository -->
   </parent>
 
-  <repositories>
+ <repositories>
     <!-- Including for fontawesome -->
     <repository>
-      <id>spring-repo</id>
-      <name>Spring Repository</name>
-      <url>https://repo.spring.io/plugins-release/</url>
+      <id>fontawesomefx-repo</id>
+      <name>FontAwesome Repository</name>
+      <url>https://bintray.com/jerady/maven/FontAwesomeFX</url>
     </repository>
   </repositories>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -31,6 +31,15 @@
...
   </parent>
 
+  <repositories>
+    <!-- Including for fontawesome -->
+    <repository>
+      <id>spring-repo</id>
+      <name>Spring Repository</name>
+      <url>https://repo.spring.io/plugins-release/</url>
+    </repository>
+  </repositories>
```
JSchraff: shoul use official binaries at https://bintray.com/jerady/maven/FontAwesomeFX  think spring repo is just a link to that but better save than sorry

```suggestion
 <repositories>
    <!-- Including for fontawesome -->
    <repository>
      <id>fontawesomefx-repo</id>
      <name>FontAwesome Repository</name>
      <url>https://bintray.com/jerady/maven/FontAwesomeFX</url>
    </repository>
  </repositories>
```

Death111: Can you please have a look on if we can tell mvn that this repo should just be used for the fontawesome libs?

JSchraff: this repository is used only if dependency is not available in cxentral maven repository

PR #49, opened 2020-03-13
Feature/#38 remember last position on screen

JSchraff:
saves position on the screen


src/main/resources/layouts/settings.fxml
```diff
@@ -223,6 +223,41 @@
                      </children>
                   </Group>
                </children>
+            </VBox>
+                        <VBox styleClass="menuBorder">
```
Death111: seems not to be formatted correctly?

src/main/resources/layouts/settings.fxml
```diff
@@ -223,6 +223,41 @@
...
+                        <VBox>
+                           <children>
+                              <HBox spacing="10.0">
+                                 <children>
+                                    <Label fx:id="hotkeyLabel1" text="Position">
+                                       <font>
+                                          <Font name="Open Sans Bold" size="12.0" />
+                                       </font>
+                                    </Label>
+                                    <Label fx:id="globalKeyloggerLabel" alignment="TOP_LEFT" contentDisplay="TOP" disable="true" prefHeight="17.0" prefWidth="120.0" text="(global keylistener!)">
```
Death111: why do we have this label here

src/main/java/de/doubleslash/keeptime/Main.java
```diff
@@ -233,6 +245,69 @@ private void initialisePopupUI(final Stage primaryStage) throws IOException {
...
+               return;
+            }
+
+            final Settings newSettings = new Settings(model.hoverBackgroundColor.get(), model.hoverFontColor.get(),
+                  model.defaultBackgroundColor.get(), model.defaultFontColor.get(), model.taskBarColor.get(),
+                  model.useHotkey.get(), model.displayProjectsRight.get(), model.hideProjectsOnMouseExit.get(),
+                  model.windowPositionX.get(), model.windowPositionY.get(), model.screenHash.get(),
+                  model.saveWindowPosition.get(), model.remindIfNotesAreEmpty.get());
+
+            if (observable.equals(primaryStage.xProperty())) {
```
Death111: should extract all the position logic into own class

src/main/java/de/doubleslash/keeptime/model/Model.java
```diff
@@ -136,6 +142,28 @@ public Project getIdleProject() {
       return allProjects;
    }
 
+   public double getAbsolutePositionX() {
+      final Screen toDisplay = getScreenWithHashOrPrimary(screenHash.get());
```
Death111: should extract all the position logic into own class

src/main/java/de/doubleslash/keeptime/model/Settings.java
```diff
@@ -59,6 +59,38 @@
 
    private boolean hideProjectsOnMouseExit;
 
+   private double windowPositionX;
```
Death111: may should start putting fields belonging together into own class like here (x,y,hash)

src/main/java/de/doubleslash/keeptime/model/Settings.java
```diff
@@ -59,6 +59,38 @@
...
+      this.hideProjectsOnMouseExit = hideProjectsOnMouseExit;
+      this.windowPositionX = windowPositionX;
+      this.windowPositionY = windowPositionY;
+      this.screenHash = screenHash;
+      this.saveWindowPosition = saveWindowPosition;
+      this.remindIfNotesAreEmpty = remindIfNotesAreEmpty;
+
+   }
+
    private boolean remindIfNotesAreEmpty;
```
Death111: field should not be between methods, but at the top

src/main/resources/db/migration/V1_1_3__added_screen_settings.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE settings ADD COLUMN window_positionx DOUBLE NOT NULL DEFAULT 0.5;
+ALTER TABLE settings ADD COLUMN window_positiony DOUBLE NOT NULL DEFAULT 0.5;
+ALTER TABLE settings ADD COLUMN screen_hash INT NOT NULL DEFAULT 0;
+ALTER TABLE settings ADD COLUMN save_window_position BOOLEAN NOT NULL DEFAULT false
```
Death111: ; missing - or not consistent with other lines

Death111: also 'formatting' is not like in other migrations

src/main/java/de/doubleslash/keeptime/Main.java
```diff
@@ -233,6 +245,69 @@ private void initialisePopupUI(final Stage primaryStage) throws IOException {
...
+         primaryStage.setX(model.getAbsolutePositionX());
+         primaryStage.setY(model.getAbsolutePositionY());
+      }
+
+      final ChangeListener<Number> posChange = new ChangeListener<Number>() {
+
+         @Override
+         public void changed(final ObservableValue<? extends Number> observable, final Number oldValue,
+               final Number newValue) {
+            if (!model.saveWindowPosition.get()) {
```
Death111: need to test if the Screen logic also works on linux

src/main/resources/db/migration/V0_0_1__init.sql
```diff
@@ -1,4 +1,4 @@
- CREATE TABLE IF NOT EXISTS project
+CREATE TABLE IF NOT EXISTS project
```
Death111, line 1: did you even try this? you cannot change migration files as the checksum will not be the same "Validate failed: Migration checksum mismatch for migration version 0.0.1"
should just change the new migration in this feature, which was not applied by everyone using develop-branch

src/main/java/de/doubleslash/keeptime/common/ScreenPosManager.java
```diff
@@ -0,0 +1,117 @@
...
+import javafx.stage.Screen;
+import javafx.stage.Stage;
+
+public class ScreenPosManager {
+
+   Stage stage;
+   Model model;
+   Controller controller;
+
+   public ScreenPosManager(final Stage stage, final Model model, final Controller controller) {
```
Death111, line 17: class should have just one concern, handling screen coordinates. Adding dependecies to other components makes it complex and hard to understand.
it should not be the concern of the class to e.g. update a model. Separation of Concerns
it should only be responsible to make crazy screen calculations. Single Responsibility

Other classes can use the functionaly of the class.

src/main/java/de/doubleslash/keeptime/common/ScreenPosManager.java
```diff
@@ -0,0 +1,117 @@
...
+import de.doubleslash.keeptime.model.Model;
+import de.doubleslash.keeptime.model.Settings;
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.ObservableValue;
+import javafx.stage.Screen;
+import javafx.stage.Stage;
+
+public class ScreenPosManager {
+
+   Stage stage;
```
Death111, line 13: no modifiers

Death111:
I need to test with multiple screens, also disconnecting them at runtime and see if it works correctly.

src/main/java/de/doubleslash/keeptime/Main.java
Death111: see new comment in new review

mplieske:
@JSchraff @Death111 
I tested this feature in various ways and it seems to be verry stable :).

mplieske:
I have found a bug and it could be from this feature. I have setup my KeepTime with this version. Setting for save window position is enabled. After a reboot from a windows update, my KeepTime starts itself from "shell:startup" (the autostart folder, not the taskmanagers autostart) and aparently is not on any screen, but on the taskbar.

Death111:
if minimized position will be set to -32000 for x and y
22:53:06.240 [JavaFX Application Thread] INFO  de.doubleslash.keeptime.Main - Pos has changed -32000.0
22:53:06.269 [JavaFX Application Thread] INFO  de.doubleslash.keeptime.Main - Pos has changed -32000.0
opening the application again will put it there making it not seeable

src/main/java/de/doubleslash/keeptime/Main.java
```diff
@@ -233,6 +245,51 @@ private void initialisePopupUI(final Stage primaryStage) throws IOException {
...
+               poshelper.setAbsoluteY(newValue.doubleValue());
+            }
+
+            final Settings newSettings = new Settings(model.hoverBackgroundColor.get(), model.hoverFontColor.get(),
+                  model.defaultBackgroundColor.get(), model.defaultFontColor.get(), model.taskBarColor.get(),
+                  model.useHotkey.get(), model.displayProjectsRight.get(), model.hideProjectsOnMouseExit.get(),
+                  poshelper.getProportionalX(), poshelper.getProportionalY(), poshelper.getScreenHash(),
+                  model.screenSettings.saveWindowPosition.get(), model.remindIfNotesAreEmpty.get());
+
+            controller.updateSettings(newSettings);
```
Death111, line 284: I'm not too confident that saving this to the db on each move is good. If you drag and drop the app there are loads of events (also x and y are separate events).
Guess we should just save it once when shutting down, and later also when autoSave is merged..

Death111: implemented to just save on shutdown

src/main/java/de/doubleslash/keeptime/Main.java
Death111: when tested on linux, i will merge it

src/main/java/de/doubleslash/keeptime/common/ScreenPosHelper.java
```diff
@@ -1,94 +1,109 @@
...
+      if (!isPositionValid()) {
+         final Screen screen = getScreenWithHashOrPrimary(this.screenHash);
+         final Rectangle2D screenBounds = screen.getBounds();
+         LOG.warn(
+               "Position is not in the range of the screen. Screen (hash '{}') range '{}'. Calculated positions '{}'/'{}'. Setting to '0'/'0'.",
+               screen.hashCode(), screenBounds, absoluteX, absoluteY);
+         proportionalX = 0;
+         proportionalY = 0;
+         absoluteX = 0;
+         absoluteY = 0;
```
JSchraff, line 74: would set proportional to 0.5 and then calculate the absolute to place in the middle  of the screen

JSchraff:
> @JSchraff @Death111
> I tested this feature in various ways and it seems to be very stable :).
> 
> List of tests I made:
> 
> * drag and restart application
> * drag to display B, remove display B
> * drag to display A, close application, remove display B, start application
> * drag to display B, close application, remove display B, start application
> * drag on display A, close application, add display B, start application
> * drag on display A, shutdown system, boot system, start application
> * drag on display A, close application, shutdown system, add display B, boot system, start application
> * drag on display A, shutdown system, add dispay B, boot system, start application
> * drag to display B, shutdown system, remove display B, boot system, start application
> * set display As resolution to 800x600, drag to (bottom right corner) display B (1080p), shutdown system, remove display B, boot system, start application

test 6,8,9 does not work anymore because of 6313a59 schould be mostly solved with autosave #53

JSchraff:
tested on ubuntu working with one and two monitors did not test edgecases.

PR #51, opened 2020-03-16
implemented selection of first project

JSchraff:
implemented selection of first project if no Project selected but enter pressed

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
@@ -241,10 +241,11 @@ public class ManageWorkController {
       });
 
       // on
-      projectComboBox.setOnKeyReleased((ke) -> {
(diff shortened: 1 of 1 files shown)


src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -239,6 +240,16 @@ public void changed(final ObservableValue<? extends String> observable, final St
 
       });
 
+      // on
+      projectComboBox.setOnKeyReleased((ke) -> {
+         if (ke.getCode() == KeyCode.ENTER && projectComboBox.getSelectionModel().isEmpty()) {
+            if (!projectComboBox.getItems().isEmpty()) {
+               projectComboBox.getSelectionModel().selectFirst();
```
Death111: this makes the UI respond weird. its looks like an endless call loop or something, the cursor starts flickering if I select a project with "Enter"

PR #52, opened 2020-03-16
added delete functionality

JSchraff:


src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -205,6 +205,14 @@ public void editWork(final Work workToBeEdited, final Work newValuedWork) {
 
    }
 
+   public void deleteWork(final Work workToBeDeleted) {
+
```
Death111: until now every method has a LOG.info at the start.

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -257,9 +260,33 @@ public void updateItem(final LocalDate item, final boolean empty) {
 
    }
 
+   private Button createDeleteWorkButton(final Work w) {
+
+      // XXX Auto-generated method stub
```
Death111: did u not look over the MR urself?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -257,9 +260,33 @@ public void updateItem(final LocalDate item, final boolean empty) {
...
+      final Button editButton = new Button("delete");
```
Death111: editButton

PR #53, opened 2020-03-17
added quicksave Interval of 60sec

JSchraff:


JSchraff:
could add quicksaveinterval as Setting

src/main/java/de/doubleslash/keeptime/viewpopup/ViewControllerPopup.java
```diff
@@ -113,8 +112,6 @@ private void initialize() {
          }
       });
 
-      Interval.registerCallBack(() -> projectsListViewController.tick());
```
Death111: why delete the call here?
Popup will not be updated I guess?

src/main/java/de/doubleslash/keeptime/view/ViewController.java
```diff
@@ -303,6 +303,8 @@ private void initialize() {
 
          mainColorTimeLine.update(model.getSortedPastWorkItems(), controller.calcTodaysSeconds());
          updateTaskbarIcon(currentWorkSeconds);
+
+         projectsListViewController.tick();
```
JSchraff: added tick to main ui update interval

Death111: duplicate tick now here see#302

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -51,44 +53,59 @@
...
+   }
+
+   public void saveCurrentWork(final long minusSeconds) {
+      final Work currentWork = model.activeWorkItem.get();
+
+      if (currentWork == null) {
+         return;
+      }
+      final LocalDateTime now = dateProvider.dateTimeNow().minusSeconds(minusSeconds);
+      final LocalDate dateNow = now.toLocalDate();
```
Death111, line 87: duplicate variables - pass them instead of minusSeconds

src/main/java/de/doubleslash/keeptime/controller/Controller.java
```diff
@@ -51,44 +53,59 @@
...
+      final String time = DateFormatter
+            .secondsToHHMMSS(Duration.between(currentWork.getStartTime(), currentWork.getEndTime()).getSeconds());
+
+      LOG.info("Saving Work from '{}' to '{}' ({}) on project '{}' with notes '{}'", currentWork.getStartTime(),
+            currentWork.getEndTime(), time, currentWork.getProject().getName(), currentWork.getNotes());
+
+      // Save in db
+      model.getWorkRepository().save(currentWork);
+
       if (currentWork != null && !dateNow.isEqual(currentWork.getCreationDate())) {
```
Death111, line 103: this logic should not be in the save work logic.
also this will confuse the tool now, if a work was started on day A and now is day B, it will remove the active item from pastWorkItems, but it is still active. need to be rethinked how this is handeled. Guess this is solved if you move it back to changeProject

Death111:
Using this version now for ~2weeks and working fine. Additional setting could be created like you mentioned, but is not needed from my point of view.

PR #58, opened 2020-03-25
Bugfix/edit work filter uses all projects

Death111:
I did not change the referenced projects in this call. only in the initial phase


Death111:
wtf - wrong base branch it seems

PR #67, opened 2020-04-28
loading calenderwidget with update

JSchraff:


Death111:
Hey @JSchraff ,

what does this fix again? Does this only fix when editing a work item to put it to another day, that the day in the calender is enabled/disabled? (which it currently does not do, only when changing the day in the calender, the update is applied)

Or did I miss something else?

Best
Nico

JSchraff:
no that is correct, that is all

PR #72, opened 2020-07-20
extracted ProjectSelectionBox Logic into its own class 

JSchraff:
extracted ProjectSelectionBox Logic into its own class  to enable reuseability


JSchraff:
could be beneficial to extract the searching logic into its own class independent of Project, but cannot find a direct usecase, that justifys the overhead of generalisation.

Death111:
this would be interesting for the #178 feature which uses a lot of comboboxes. I was not able to get it work :p
It had issues when using in a ListView (which reuses comboxes when scrolling.

Issue #79, opened 2021-03-21
Wrong time calculation when startTime after endTime (manual edit)


Death111:
![image](https://user-images.githubusercontent.com/5772511/111920839-91722500-8a91-11eb-8194-cf06b4e086c9.png)

The item with 33 minutes has properties: startTime=2021-03-02T16:34:33.525, endTime=2021-03-02T16:00:37

JSchraff:
with #82 it is no longer possible to create Dates with negative duration. #80 will not be merged for now. It is still necessary to unify the use of DateFormatter and Duration. It should not be obfuscating the existence of negative durations, because this could lead to missunderstandings and hampers the ability to spot Problems.
Therefore it should be formatted to show the negative Time everywhere correctly if somehow a negative Time gets created, or in future it is necessary to support negative work times.

Death111:
I also noticed that "absolute time" is calculated wrongly when there are overlapping times.
- if I have two work items and edit them to start and stop at 14-15 o'clock.
   - Expected: Absolute time 1 hours
   - Actual: Absolute time 2 hours

It can also be discussed on what is expected behavior here. As if both items would be "work" we would have absolute time 1h and work time displayed as 2 h - which is also confusing

PR #80, opened 2021-03-21
Bugfix/#79 negative time calculation

Death111:
Fixes issue where time was calculated wrong in report for the whole project, when work items exist which have startTime after endTime.
Issue was in Controller.calcSeconds but I also adapted logic at the other places


Death111:
Maybe it would be more obvious for the user when the negative time is shown in the work items. If this is done have a look at the report timeline, as this will most likely not work with negative times.

Instead of hiding the wrong start end time selection by just using the abs() method (like it was fixed now).

JSchraff (commented):
see: https://github.com/doubleSlashde/KeepTime/issues/79#issuecomment-819466662

PR #81, opened 2021-03-21
Feature/ipte 22 export h2 database to sql script

Death111:


Death111:
Hint: When creating new release, we need to include steps how to move old h2 database to new folder '/db/x.x.x/'

I need to update ReleaseNotes to include changes:
- database path includes database version
- Added export button in settings dialog, to export to .sql (for later import coming with next release)
- Logging build related infos on startup

JSchraff (commented):
I see some Issues concerning the Import of the generated sql, see comment in SettingsController.

Even after Drop Statements is the execution of the  flyway migration after import necessary.
This might run into issues if a flyway migration was run inbetween export and Import if tables got renamed and if run again table already exists (changed tables will not be dropped).

JSchraff (approved):

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -1,334 +1,334 @@
-// Copyright 2019 doubleSlash Net Business GmbH
```
Death111, line 1: you reformatted the whole file content

Death111: fixed

PR #82, opened 2021-04-08
Bugfix/#79 prevent negative work times

JSchraff:


Death111 (approved):
I changed the logic to disable the OK button. Other than that this is fine.

PR #94, opened 2022-09-16
added code security scanner

sesturm:


Death111:
I like it.
FYI we are planning to:
- Upgrade to Java 11
- Get rid of FontAwesome dependency (as it is no more available)
- Move to github action for CI
  - also wanted to integrate [sonarqube-scan](https://github.com/marketplace/actions/sonarqube-scan)

PR #97, opened 2022-10-13
Feature/#92 word notes dialog

DavidDamke:


src/main/java/de/doubleslash/keeptime/view/ProjectsListViewController.java
```diff
@@ -169,27 +169,38 @@ private void changeProject(final Project newProject, final long minusSeconds) {
...
-               currentWork.setNotes(result.get());
-            } else {
-               // cancel pressed
-               return;
-            }
+         if (currentWork != null && currentWork.getNotes().isEmpty() ) {
+            if (showNoNoteSelected(currentWork)) return;
+         }
+      }
+      if (model.remindIfNotesAreEmptyIsWork.get()) {
```
Death111, line 176: the _remindIfNotesAreEmpty_ and _remindIfNotesAreEmptyIsWork_ should work connected. Both can be true. In theory the new option only makes sense if the existing one is already active (but we do not need to add logic for this).

PseudoCode:
if (remindIfNotesAreEmpty && notesAreEmpty)
    if (!remindIfNotesAreEmptyIsWork || isWork)
       promptForNote

src/main/resources/db/migration/V1_1_6__renamed_note_reminder_setting_only_work.sql
```diff
@@ -0,0 +1,2 @@
+ ALTER TABLE settings
+  RENAME COLUMN empty_note_reminder_only_work TO empty_note_reminder_is_work
```
Death111, line 2: as you are creating one new version with the new feature - one .sql script is sufficient. Only you locally have the issue that the column has the "old" name. But locally you can just delete the database so it is recreated with the "new" name.
So you can delete 1.1.6 and just use the correct name in 1.1.5.

If a user installs the new version, he never knew the old field name.

src/main/resources/layouts/settings.fxml
```diff
@@ -283,6 +283,35 @@
...
+                                    <Label text="Note Reminder">
+                                       <font>
+                                          <Font name="Open Sans Bold" size="12.0" />
+                                       </font>
+                                    </Label>
+                                 </children>
+                              </HBox>
+                              <HBox spacing="10.0">
+                                 <children>
+                                    <CheckBox fx:id="emptyNoteReminderCheckBoxIsWork" mnemonicParsing="false" text="Ask for notes when switching from project marked as work (if empty)" wrapText="true" HBox.hgrow="NEVER">
```
Death111, line 303: the new option name (and variable names) are not so clear to be.
If the option is not selected KeepTime should ask for a note for each project (if note reminder is on). If true, only for the ones which are work.

e.g. renaming "remindIfNotesAreEmptyIsWork" to "remindIfNotesAreEmptyOnlyForIsWork" would make it a little clearer what it means. maybe there is even better name.

Death111 (changes requested):

Death111 (approved):

PR #102, opened 2022-10-24
Feature/import sql to database

DavidDamke:


src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -239,6 +243,56 @@ private void initialize() {
...
+            }
+
+            final FileChooser fileChooser = new FileChooser();
+            fileChooser.setTitle("Open Exported SQl Script");
+            File file = fileChooser.showOpenDialog(thisStage);
+            if (file == null) {
+               LOG.info("User canceled import.");
+               return;
+            }
+            LOG.info(file.toString());
```
Death111, line 268: add some context to the log message

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -239,6 +243,56 @@ private void initialize() {
...
+
+            final String url = applicationProperties.getSpringDataSourceUrl();
+            final String username = applicationProperties.getSpringDataSourceUserName();
+            final String password = applicationProperties.getSpringDataSourcePassword();
+            RunScript.execute(url, username, password, file.toString(), Charset.defaultCharset(), true);
```
Death111, line 273: [H2 Migration to v2](https://www.h2database.com/html/migration-to-v2.html) states that we may need to specify `FROM_1X flag`. Do we need that?

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -239,6 +243,56 @@ private void initialize() {
...
       });
    }
+   private void initImportButton(){
+      LOG.debug("Initialize importButton.");
+      importButton.setOnAction(event ->{
+
+         try {
+            Alert confirmationAlert = new Alert(AlertType.CONFIRMATION , "", ButtonType.YES, ButtonType.NO);
+            confirmationAlert.setTitle("Import");
+            confirmationAlert.setHeaderText("Do you want to Override current Data ?");
```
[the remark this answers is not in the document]
DavidDamke: Added the suggested text to the alert content

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -239,6 +243,56 @@ private void initialize() {
...
+            confirmationAlert.showAndWait();
+
+            if(confirmationAlert.getResult()==ButtonType.NO){
+               LOG.info("User canceled import");
+               return;
+            }
+
+            final FileChooser fileChooser = new FileChooser();
+            fileChooser.setTitle("Open Exported SQl Script");
+            File file = fileChooser.showOpenDialog(thisStage);
```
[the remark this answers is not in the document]
DavidDamke: Added filter

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
DavidDamke: As far as I tested it it worked without the tag.

Death111: the docs state "Use this clause when you import script that was generated by H2 1.4.200 or an older version into more recent version.". As this is the case for us we should set the flag.

Later (if another upgrade is needed) we may need to validate the version we want to import (to set the flag if we come from 1.x and dont set it from 2.x).

DavidDamke: Added the option

Death111:
Can you also update the README migrate-from-old-version section to include the new export/import mechanism.

We should keep the current instructions with a note that these are needed from an older version (that are the needed steps to migrate from a previous version to the currently released one). The new instructions (you are adding) are only valid if they already use the latest released version. If they use a previous version they first need to migrate to the current version, do the export and then upgrade to the to be released version.

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
Death111: can you add a // TODO above the RunScript, to add an option at the next release to make the "FROM_1X flag" configureable. E.g. if we upgrade (in the release after) from H2 version 2.x to 2.x we must not set the "FROM_1X flag".

Death111 (approved):
On linux the Import/Export button crashes the VM - seems to be caused by the fileChooser.
We should add instruction on how to import/export on a linux system.
We will have a look in a different issue regarding this (as import is also already broken).

Issue #108, opened 2022-11-07
Import export broken in linux - vm crashes


Death111:
On linux the Import/Export button crashes the VM - seems to be caused by the fileChooser.

PR #109, opened 2022-11-08
import and export works with jnativehook version 2.1.0

DavidDamke:
Version 2.03 and 2.1.0 do not work on Linux. System crashes after filechooser is being executed


pom.xml
```diff
@@ -71,7 +71,7 @@
     <dependency>
       <groupId>com.1stleg</groupId>
       <artifactId>jnativehook</artifactId>
-      <version>2.1.0</version>
+      <version>2.0.2</version>
```
Death111, line 74: I would not like to downgrade the dependency.

Can you please validate the error also happens, if we would not create an instance of the jnativehook on linux? I think currently we create one on every OS but do not "register listeners" for linux based systems. Maybe not creating an instance already fixes the issue?

DavidDamke: if you  do not "register listeners" for linux it works on Java 8 +11  with the latest jnativehook version

Death111:
This branch fixes #108

src/main/java/de/doubleslash/keeptime/App.java
```diff
@@ -218,10 +219,10 @@ private void initialisePopupUI(final Stage primaryStage) throws IOException {
       LOG.debug("Initialising popup UI.");
 
       globalScreenListener = new GlobalScreenListener();
```
Death111, line 221: My idea was to put the if(!OS.isLinux()) switch to the creation of the Listener. If we are on linux, we do not need to create an instance of it :)
- So the change in the GlobalScreenListener class would not be needed.
- The changes in this class would need to validate globalScreenListner!=null insted of the linux switch.

src/test/java/de/doubleslash/keeptime/common/DateFormatterTest.java
```diff
@@ -28,6 +28,7 @@
 public class DateFormatterTest {
 
    @Test
+   @Disabled
```
Death111, line 31: why do we need to disable it now?

Death111 (changes requested):

src/test/java/de/doubleslash/keeptime/common/DateFormatterTest.java
DavidDamke: I thought it got executed although the class was Disabled.  But I tested it and now it works again

src/main/java/de/doubleslash/keeptime/App.java
DavidDamke: reverted the changes in GlobalScreenListener

Death111 (approved):

PR #107, opened 2022-11-04
Feature/logo

DavidDamke:

PR #112, opened 2022-11-14
Bug/dialog windows hidden and dont show full text

DavidDamke:
Bug Fixed: If you press the About Button then the Import/Export Dialog would be behind the Settings dialog.  
Bug Fixed: Only Part of the Import/Export  Dialog text was shown in Linux


src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -407,8 +411,8 @@ private void initImportButton() {
             Alert informationDialog = new Alert(AlertType.INFORMATION);
             informationDialog.setTitle("Import done");
             informationDialog.setHeaderText("The data was imported.");
-            informationDialog.setContentText(
-                  "KeepTime will now be CLOSED! You have to RESTART it again to see the changes");
+            informationDialog.getDialogPane().setContent(new Label("KeepTime will now be CLOSED!\n" + "You have to RESTART it again to see the changes"));
+            informationDialog.initOwner(thisStage);
```
Death111, line 415: can you retest the behavior after we now merged #107 where we removed the "initOwner" of most of the dialogs?

Death111 (approved):

PR #113, opened 2022-11-15
Bug/hide projects not supported on linux

DavidDamke:
The option Hide Projects on mouse leave does not work on Linux. Therefore it is now disabled


src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -190,16 +191,6 @@ private void initialize() {
 
                alert.showAndWait();
             }
-            if (!displayProjectsRightCheckBox.isSelected() && hideProjectsOnMouseExitCheckBox.isSelected()) {
```
Death111, line 193: can you add logic here, to set the value of hideProjects to false, if it is true. Maybe some users of a previous version have it active. We need to deactivate it then on settings save :)

DavidDamke: Option is disabled and on save it get set to false 👍

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -190,15 +191,9 @@ private void initialize() {
...
+            if(hideProjectsOnMouseExitCheckBox.isSelected()){
                hideProjectsOnMouseExitCheckBox.setSelected(false);
-               final Alert warning = new Alert(AlertType.WARNING);
-               warning.setTitle("Warning!");
-               warning.setHeaderText("No Linux Support");
-               warning.getDialogPane().setMinHeight(Region.USE_PREF_SIZE);
-               warning.setContentText(
-                     "The project list on the left side has no Linux support if projects should be hidden. Disabling hiding of project list.");
-               warning.showAndWait();
+               hideProjectsOnMouseExitCheckBox.setDisable(true);
```
Death111, line 196: we don't need to disable it (again)? it already is disabled from the start?

Death111: removed the line

Death111 (approved):

PR #115, opened 2022-11-17
Bug/project colors in report on wrong position

DavidDamke:


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -156,6 +156,7 @@ protected void updateItem(final TableRow item, final boolean empty) {
                      label.setUnderline(item.isUnderlined());
                      label.setTooltip(new Tooltip(notes));
                      this.setGraphic(label);
+                     workTableTreeView.refresh();
```
Death111, line 159: can you try to move the Circle into this update method? so that for ProjectRows the circle is set as graphic and the project name as text.
- As the text cannot be underlined (for work projects) not sure if we want to indicate work items another way
   - maybe we underline the time of the project instead?

DavidDamke: Moved the text and the circle to the Update Method. 
For the work maybe add an extra column ?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -152,10 +153,11 @@ protected void updateItem(final TableRow item, final boolean empty) {
                      setText(null);
                   } else {
                      final String notes = item.getNotes();
-                     final Label label = new Label(notes.isEmpty() ? EMPTY_NOTE : notes);
-                     label.setUnderline(item.isUnderlined());
-                     label.setTooltip(new Tooltip(notes));
-                     this.setGraphic(label);
+                     final Text text = new Text(notes.isEmpty() ? EMPTY_NOTE : notes);
+                     this.setText(text.getText());
```
Death111, line 157: we do not need the "Text" class when in the end you just pass a String to the setText method. A normal String is sufficient ;)

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -152,10 +153,11 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+
+                     final Circle circle = new Circle(6, item.getProjectColor());
```
Death111, line 159: this is not right or? we can only set a graphic if we also have a color? Creating a color with "null" as color (which happens for the notes) is not okay.
if we don't have a color we need to set graphic to 'null'.

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
Death111: lets just put a border on the circle for now - when it is work
circle.setStroke(Color.BLACK);

DavidDamke: I underlined the Time column to indicate that it is a work item. I can also add thie cirlce border

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Graphic no only get set if the  item.getProjectColor is not null

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -176,7 +182,35 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+         @Override
+         public TreeTableCell<TableRow, String> call(TreeTableColumn<TableRow, String> tableRowStringTreeTableColumn) {
+
+            return new TreeTableCell<TableRow,String>(){
+
+               @Override
+               protected void updateItem(String s, boolean b) {
+                  super.updateItem(s, b);
+                  Label l1 = new Label(s);
+                  if(b){
```
Death111, line 195: if should look different (from the javadoc) and also reset Text to null

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -176,7 +182,35 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+      timeSumColumn.setCellFactory(new Callback<TreeTableColumn<TableRow, String>, TreeTableCell<TableRow, String>>() {
+         @Override
+         public TreeTableCell<TableRow, String> call(TreeTableColumn<TableRow, String> tableRowStringTreeTableColumn) {
+
+            return new TreeTableCell<TableRow,String>(){
+
+               @Override
+               protected void updateItem(String s, boolean b) {
+                  super.updateItem(s, b);
+                  Label l1 = new Label(s);
```
Death111, line 194: why create a label with a possibly null value? and later check 'l1.getText()!=null'. should make sure that s is not null and only then create a label with it

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -176,7 +182,35 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
       final TreeTableColumn<TableRow, String> timeSumColumn = new TreeTableColumn<>("Duration");
-      timeSumColumn.setCellValueFactory(new TreeItemPropertyValueFactory<TableRow, String>("timeSum"));
+      timeSumColumn.setCellFactory(new Callback<TreeTableColumn<TableRow, String>, TreeTableCell<TableRow, String>>() {
+         @Override
+         public TreeTableCell<TableRow, String> call(TreeTableColumn<TableRow, String> tableRowStringTreeTableColumn) {
+
+            return new TreeTableCell<TableRow,String>(){
+
+               @Override
+               protected void updateItem(String s, boolean b) {
```
Death111, line 192: please rename the parameters to what they actually are. not clear what "s" or "b" should represent

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
Death111: can leave it like this then. just thought we could save the hastle with custom cell renderer - but as it is already implemented now we can keep it

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: renamed the parameters

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: changed it as in https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Cell.html#updateItem-T-boolean-

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Added your suggestion

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -176,7 +164,37 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+               protected void updateItem(String timeString, boolean empty) {
+                  super.updateItem(timeString, empty);
+
+                  if (timeString == null || empty) {
+                     this.setGraphic(null);
+                     this.setText(null);
+                  } else {
+                     this.setGraphic(new Label(timeString));
+
+                     for (String workTime : workTime) {
```
Death111, line 183: why do we need a for-loop here? not sure what this logic does actually. It should set the "timeString" as underline I guess? We should not reference external variables. Only the ones from updateItem method.

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: I need the external variable to check if a time is a work time or a regular. Its the only way I found to find out if a time belongs to a work item.

DavidDamke: The forloop is needed to underline all working times

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -122,6 +104,8 @@ public class ReportController {
 
    private final TreeItem<TableRow> rootItem = new TreeItem<>();
 
+   private static List<String> workTimeList = new ArrayList<>();
```
Death111, line 107: can be deleted

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -175,8 +163,36 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+                  super.updateItem(workItem, empty);
+
+                  if (workItem == null || empty) {
+                     this.setGraphic(null);
+                     this.setText(null);
+                  } else {
+                     this.setGraphic(new Label(workItem.getTimeSum()));
+
+                     if(workItem.isUnderlined()){
+                        Label workLabel = new Label(workItem.getTimeSum());
```
Death111, line 184: looks complicated. Why not just
```
Label worlLabel = new Label(workItem.getTimeSum());
workLabel.setUnderline(workItem.isUnderlined());
this.setGraphic(workLabel);
this.setText(null);
```

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -220,6 +236,10 @@ private void updateReport(final LocalDate dateToShow) {
 
          final long projectWorkSeconds = controller.calcSeconds(onlyCurrentProjectWork);
 
+         if(project.isWork()){
+            workTimeList.add(DateFormatter.secondsToHHMMSS(projectWorkSeconds));
```
Death111, line 240: not needed?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: deleted

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: deleted

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Implemented it :)

PR #125, opened 2022-11-24
Enhancement/report has button to expand and close

### LINKED ISSUE #119: Report controller add "collapse/expand" all feature
In the report controller add functionality to allow the user to expand or collapse all projects.

Idea:
- a button (or two?) to collapse or expand all projects

Optional:
- the state is saved when switching days

DavidDamke:
#119

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/ReportController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ReportController.java
@@ -215,7 +215,7 @@ public class ReportController {
       expanded = !expanded;
    }
 
-   private void expand(boolean expand){
+   private void expandAll(boolean expand){
       for (int i=0; i<rootItem.getChildren().size(); i++){
          rootItem.getChildren().get(i).setExpanded(expand);
       }
--- a/src/main/java/de/doubleslash/keeptime/view/ReportController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ReportController.java
@@ -140,7 +140,7 @@ public class ReportController {
 
       colorTimeLine = new ColorTimeLine(colorTimeLineCanvas);
 
-      expandCollapseButton.setOnMouseClicked(event ->collapseExpandTable());
+      expandCollapseButton.setOnMouseClicked(event ->toggleCollapseExpandReport());
       initTableView();
    }
 
@@ -203,14 +203,12 @@ public class ReportController {
    private void toggleCollapseExpandReport(){
 
       if(expanded){
-         //CollapseAll
-         expand(false);
-         expandCollapseButton.setText("Expand");
+         expandAll(false);
+         expandCollapseButton.setText("Expand Report list");
 
       }else {
-         //ExpandAll
-         expand(true);
-         expandCollapseButton.setText("Collapse");
+         expandAll(true);
+         expandCollapseButton.setText("Collapse Report list");
       }
       expanded = !expanded;
    }
--- a/src/main/java/de/doubleslash/keeptime/view/ReportController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ReportController.java
@@ -26,7 +26,6 @@ import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import javafx.scene.control.skin.DatePickerSkin;
-import org.apache.commons.logging.Log;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -110,7 +109,7 @@ public class ReportController {
    private Canvas colorTimeLineCanvas;
 
    @FXML
-   private Button expandColapseButton;
+   private Button expandCollapseButton;
 
    private static final Logger LOG = LoggerFactory.getLogger(ReportController.class);
 
@@ -126,7 +125,7 @@ public class ReportController {
 
    private final TreeItem<TableRow> rootItem = new TreeItem<>();
 
-   private static boolean pressed= false;
+   private static boolean expanded = true;
 
    @Autowired
    public ReportController(final Model model, final Controller controller) {
@@ -141,7 +140,7 @@ public class ReportController {
 
       colorTimeLine = new ColorTimeLine(colorTimeLineCanvas);
 
-      expandColapseButton.setOnMouseClicked(event ->collapseExpandTable());
+      expandCollapseButton.setOnMouseClicked(event ->collapseExpandTable());
       initTableView();
    }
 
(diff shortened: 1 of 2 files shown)


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -194,6 +198,18 @@ protected void updateItem(final TableRow item, final boolean empty) {
       rootItem.setExpanded(true);
    }
 
+   private void collapseExpandTable(){
+      for (int i=0; i<rootItem.getChildren().size(); i++){
```
Death111, line 202: the button should expand all or collapse all. it should not 'stupidly' toggle the current status. would not be to useful.
with your logif currently e.g.  you collapsed one manualy (all others are still expanded). you press the button. now the collapsed one will expand and the others collapse

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -194,6 +198,18 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+
+         if( rootItem.getChildren().get(i).isExpanded()){
+            rootItem.getChildren().get(i).setExpanded(false);
+            expandColapseButton.setText("Expand");
+         }else {
+            rootItem.getChildren().get(i).setExpanded(true);
+            expandColapseButton.setText("Collapse");
```
Death111, line 209: setting the text of the same button multiple times in a for loop will (in the end) set the text to the status of the last element in the loop. does not make so much sense.

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Button now switches between the states "expand all" and "collapse all". 
The problem is now if you manually collapse all then the button will still have the text "collapse". If that happens you have to press the button twice to expand them all again.

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: is not in the loop anymore

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -26,6 +26,7 @@
 import java.util.stream.Collectors;
 
 import javafx.scene.control.skin.DatePickerSkin;
+import org.apache.commons.logging.Log;
```
Death111, line 29: not needed

src/main/resources/layouts/report.fxml
```diff
@@ -77,6 +78,16 @@
                               </Label>
                            </children>
                         </HBox>
+                        <HBox prefHeight="26.0" prefWidth="210.0">
+                           <children>
+                              <Label prefHeight="26.0" prefWidth="137.0" text="Collapse/Expand: ">
```
Death111, line 83: not sure if we need a label "Collapse/Expand" when the button has the same info "Collapse"?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -194,6 +201,25 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+   private void collapseExpandTable(){
+
+      pressed=!pressed;
+
+      if(pressed){
+         //CollapseAll
+         for (int i=0; i<rootItem.getChildren().size(); i++){
+            rootItem.getChildren().get(i).setExpanded(false);
+         }
+         expandColapseButton.setText("Expand");
```
Death111, line 213: it is not clear to me what the state is when 'pressed' is true. It seems pressed true means the items are not expanded. the variable should be renamed so the relation to expanded/collapsed is clear(er).

Also it is not clear what the initial state is of the items - as they do not use any of the variables newly introduced. E.g. if I start the app everything is expanded, if I press the button it may want to expand again? The state should be in sync - best by using the same variables representing the state.

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -122,6 +126,8 @@ public class ReportController {
 
    private final TreeItem<TableRow> rootItem = new TreeItem<>();
 
+   private static boolean pressed= false;
```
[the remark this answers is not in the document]
DavidDamke: removed the static. Named the variable "expanded"

src/main/resources/layouts/report.fxml
DavidDamke: Removed the unnecessary label

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: removed it.

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -122,6 +125,8 @@ public class ReportController {
 
    private final TreeItem<TableRow> rootItem = new TreeItem<>();
 
+   private static boolean expanded = true;
```
Death111, line 128: can you use this variable also in
projectRow.setExpanded(true); 
instead of hardoced true?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -194,6 +200,26 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
+
+      }else {
+         //ExpandAll
+         expand(true);
+         expandCollapseButton.setText("Collapse");
+      }
+      expanded = !expanded;
+   }
+
+   private void expand(boolean expand){
```
Death111, line 218:
```suggestion
   private void expandAll(boolean expand){
```

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -194,6 +200,26 @@ protected void updateItem(final TableRow item, final boolean empty) {
...
 
+   private void collapseExpandTable(){
+
+      if(expanded){
+         //CollapseAll
+         expand(false);
+         expandCollapseButton.setText("Expand");
+
+      }else {
+         //ExpandAll
```
Death111, line 211: do we need the comments if we rename the method?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -194,6 +200,26 @@ protected void updateItem(final TableRow item, final boolean empty) {
       rootItem.setExpanded(true);
    }
 
+   private void collapseExpandTable(){
```
Death111, line 203:
```suggestion
   private void toggleCollapseExpandReport(){
```

src/main/resources/layouts/report.fxml
```diff
@@ -77,6 +78,11 @@
                               </Label>
                            </children>
                         </HBox>
+                        <HBox prefHeight="26.0" prefWidth="210.0">
+                           <children>
+                              <Button fx:id="expandCollapseButton" mnemonicParsing="false" prefHeight="25.0" prefWidth="79.0" text="Collapse" />
```
Death111, line 83: Can you have a look on a better position of the button?
image
currently it is not clear what the button will "expand"

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: deleted them

src/main/resources/layouts/report.fxml
DavidDamke: Changed the text to "Expand Report list"

Death111: That is already better. But the position is not good / the context "report list" should not be needed to be stated.
Please have a look at the basic design principals:
- https://www.youtube.com/watch?v=YkXRt9kMfzU
- https://achtung-designer.com/gestaltgesetze-die-designs-besser-machen/
They are very helpful to know (at least for the end users if they were applied) :)

DavidDamke: reworked it a bit 

grafik

Death111 (approved):

Issue #85, opened 2021-04-14
current work visible in report but changes get discarded 


JSchraff:
the current work is visible in the report after the first time the autosave is run. then it is possible to edit it which will be done.
But on the next autosave the changes get overwritten, because the activeWorkItem isn't getting updated.

Proposed Solutions:
1. remove the active Work Item from the Report, so it cannot be changed.
2. update the activeWorkItem if the workitem gets changed.

swa-ds:
I vote for option 2 (if feasible), because sometimes I want to enter notes directly after I switched to a new project.

PR #127, opened 2022-11-24
Bug/notes disappear after autosave2

DavidDamke:
Active Work Item is not editable anymore because of #85. It is still visible in the report but can only be edited when the user switches to another Work Item


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -240,6 +242,15 @@ private void updateReport(final LocalDate dateToShow) {
             workButtonBox.getChildren().add(createDeleteWorkButton(w));
             final TreeItem<TableRow> workRow = new TreeItem<>(new WorkTableRow(w, workButtonBox));
             projectRow.getChildren().add(workRow);
+            if(w.getId()==model.activeWorkItem.get().getId()){
+               workButtonBox.getChildren().remove(0,3);
```
Death111, line 246: can't you do a clear() or something like that? We want to remove all children (the number may change at some point in time).

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -26,6 +26,8 @@
 import java.util.stream.Collectors;
 
 import javafx.scene.control.skin.DatePickerSkin;
+import javafx.scene.text.Font;
+import org.apache.commons.logging.Log;
```
Death111, line 30: dont think we need this import?

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Changed it like you suggested

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Deleted it 👍

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -235,9 +236,18 @@ private void updateReport(final LocalDate dateToShow) {
 
          for (final Work w : onlyCurrentProjectWork) {
             final HBox workButtonBox = new HBox(5.0);
-            workButtonBox.getChildren().add(createCopyWorkButton(w));
-            workButtonBox.getChildren().add(createEditWorkButton(w));
-            workButtonBox.getChildren().add(createDeleteWorkButton(w));
+            if(w.getId()==model.activeWorkItem.get().getId()){
+               Label label = new Label("Active Work");
+               label.setTooltip(new Tooltip("The active work item cannot be edited as it is currently active. To edit it you need to switch to another work first."));
+               label.getTooltip().setFont(Font.font(15));
```
Death111, line 242: not sure if we need this big size for the tooltip? had it a reason? (the other tooltips we have are not this big)

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -235,9 +236,18 @@ private void updateReport(final LocalDate dateToShow) {
...
+               label.setFont(Font.font(12.0));
```
Death111, line 243: do we need to set the font size? the other labels we create do not have it set

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Thought it might be hard to read if its size 12

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: Deleted it and it looks the same :)

Issue #121, opened 2022-11-18
When editing work, automatically focus on the Notes field


Death111:
When editing a work item the notes filed should be in focus directly. Most times we want to edit notes and not the project (which is currently selected by default).

PR #128, opened 2022-11-24
Enhancement/edit work should auto focus note field

DavidDamke:
#121 Changed focus to the note field

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
@@ -109,6 +109,7 @@ public class ManageWorkController {
       setUpTimeRestriction();
 
       setProjectUpComboBox();
+
       Platform.runLater(() -> noteTextArea.requestFocus());
    }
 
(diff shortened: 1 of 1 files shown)


src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
```diff
@@ -110,7 +110,7 @@ private void initialize() {
 
       setProjectUpComboBox();
 
-      Platform.runLater(() -> projectComboBox.requestFocus());
+      Platform.runLater(() -> noteTextArea.requestFocus());
```
Death111, line 113: Was not requested: But would it be possible to move the cursor to the end of the note? may be nice :)

Death111 (approved):

src/main/java/de/doubleslash/keeptime/view/ManageWorkController.java
DavidDamke: Cursor is now at the end of the Text 👍

Death111 (approved):

PR #129, opened 2022-12-08
Feature/add dependency check in pipeline

DavidDamke:


suppressionFile.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
```
Death111, line 2: pipeline should show a warning (if possible) if a vulnerability is found. if we ignore them here - we will forget that they exist.
Let's remove the suppression file.

Death111 (changes requested):

suppressionFile.xml
DavidDamke: Removed the suppression file- 
Seams like something like allow-failure is not implemented yet https://github.com/actions/toolkit/issues/399

Death111: yes. I readded the supression as we can only fail or pass a step :/

PR #130, opened 2022-12-13
Feature/update patch version in pipeline automatically

DavidDamke:


pom.xml
```diff
@@ -32,6 +32,7 @@
   </parent>
 
   <properties>
+    <project.version>KeepTime-1.2.0-SNAPSHOT</project.version>
```
Death111, line 35: this is not a valid version.
https://maven.apache.org/pom.html#maven-coordinates

DavidDamke: I removed "KeepTime" from the version because it was doubled. But I am not sure if that was the thing you meant

Death111: yes, that was the issue.

pom.xml
```diff
@@ -32,6 +32,7 @@
   </parent>
 
   <properties>
+    <project.version>1.2.0-SNAPSHOT</project.version>
```
Death111, line 35: defining it as a property is interesting. not sure if there are disadvantages doing it this "variable" - looking fine i think

Death111 (approved):

PR #138, opened 2023-01-13
Spring 3 + Mac

Death111:
Test-branch for a Mac M1:

- Needs Java 17 as only Java 17 added support for Mac M2 (https://openjdk.org/jeps/391)
- Deactivated Unit tests as they did not run. Issue was that Controller class instantiates ([indirectly](https://github.com/doubleSlashde/KeepTime/blob/feature/macSupport/src/main/java/de/doubleslash/keeptime/common/time/Interval.java#L52)) a Timeline-Class (javafx class) which did not work in the pipeline. Not sure if I tested locally.
- Switched the isLinux() calls to isWindows() as the mac does not identify as isLinux

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>3.1.3</version>
+    <version>3.1.10</version>
     <relativePath />
     <!-- lookup parent from repository -->
   </parent>
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <java.version>17</java.version>
-    <javafx.version>20.0.1</javafx.version>
+    <javafx.version>22</javafx.version>
 
     <maven-dependency-check.version>8.0.2</maven-dependency-check.version>
     <!-- USING HTML,XML (comma-separated list) did not work with plugin version 5.1.0 -->
--- a/pom.xml
+++ b/pom.xml
@@ -129,19 +129,6 @@
       <artifactId>hamcrest-library</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.testfx</groupId>
-      <artifactId>testfx-junit5</artifactId>
-      <version>4.0.16-alpha</version>
-      <scope>test</scope>
-    </dependency>
-    <!-- https://mvnrepository.com/artifact/org.testfx/openjfx-monocle -->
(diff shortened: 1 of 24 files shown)


pom.xml
```diff
@@ -19,24 +32,11 @@
...
-  </parent>
-
   <properties>
-    <project.version>1.3.0-SNAPSHOT</project.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
+    <java.version>17</java.version>
+    <javafx.version>20.0.1</javafx.version>
```
Death111: currently just to don't forget: "JavaFX 20 Requires JDK 17 or Later"
https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-20.md#javafx-20-requires-jdk-17-or-later

If we wanted to play around with compile target version the javafx dependency would require 17 at least

pom.xml
```diff
@@ -1,16 +1,29 @@
...
+    <artifactId>spring-boot-starter-parent</artifactId>
+    <version>3.1.10</version>
+    <relativePath />
+    <!-- lookup parent from repository -->
+  </parent>
   <groupId>de.doubleslash</groupId>
   <artifactId>keeptime</artifactId>
-  <version>${project.version}</version>
-  <packaging>jar</packaging>
+  <version>2.0.0-SNAPSHOT</version>
```
Death111: version in build artifact(metainf) and filename (.zip) is not replaced like this (should use version passed from mvn package commandline)

Death111:
FileOpenHelper and BrowserHelper most likely do not work for MAC due to if/else should resovle to not supported:
https://github.com/doubleSlashde/KeepTime/blob/019bffd40628a53886a4ea2195f1ca71af7b9e52/src/main/java/de/doubleslash/keeptime/common/BrowserHelper.java#L38

src/main/java/de/doubleslash/keeptime/App.java
```diff
@@ -239,7 +243,7 @@ private void initialisePopupUI(final Stage primaryStage) throws IOException {
       final ViewControllerPopup viewControllerPopupController = loader.getController();
       viewControllerPopupController.setStage(popupViewStage);
 
-      if (!OS.isLinux()) {
+      if (OS.isWindows()) {
```
Death111, line 246: FileOpenHelper and BrowserHelper most likely do not work for MAC due to if/else should resovle to not supported:
https://github.com/doubleSlashde/KeepTime/blob/019bffd40628a53886a4ea2195f1ca71af7b9e52/src/main/java/de/doubleslash/keeptime/common/BrowserHelper.java#L38

sesturm (approved):

PR #139, opened 2023-01-13
Bug/add license header

### LINKED ISSUE #136: Add license header to all our .java files
Like: https://github.com/doubleSlashde/KeepTime/blob/master/src/main/java/de/doubleslash/keeptime/Main.java#L1-L15

DavidDamke:
fixes #136


Death111 (approved):
I may have used the current year but most of the files are already old - should be fine I guess ;)

PR #142, opened 2023-01-17
update export text

DavidDamke:


src/main/resources/layouts/settings.fxml
```diff
@@ -414,7 +414,7 @@
                                           </HBox>
                                           <HBox prefHeight="42.0" prefWidth="366.0">
                                              <children>
-                                                <Label text="Export your current Data to an Sql script. This is needed if you go from the Snapshot-1.2.0-KeepTimeto the recent release version. " wrapText="true" />
+                                                <Label text="Export your current Data to an Sql script. This is needed if you upgrade to an newer KeepTime version. " wrapText="true" />
                                              </children>
```
Death111, line 418: Could you also check the icon for the export done confirmation?
image
(not related to the issue, but would fit)

Death111 (changes requested):
its good but we may add the missing icon as well

src/main/resources/layouts/settings.fxml
DavidDamke: grafik
Added it :)

Death111 (approved):

src/main/resources/layouts/settings.fxml
```diff
@@ -414,7 +414,7 @@
                                           </HBox>
                                           <HBox prefHeight="42.0" prefWidth="366.0">
                                              <children>
-                                                <Label text="Export your current Data to an Sql script. This is needed if you go from the Snapshot-1.2.0-KeepTimeto the recent release version. " wrapText="true" />
+                                                <Label text="Export your current Data to an Sql script. This is needed if you upgrade to an newer KeepTime version. " wrapText="true" />
```
Death111, line 417:
```suggestion
                                                <Label text="Export your current Data to an Sql script. This is needed if you upgrade to a newer KeepTime version. " wrapText="true" />
```

PR #147, opened 2023-01-19
Bug/empty notes should not have semikolon

### LINKED ISSUE #146: Copied notes have prepended ; if first work item is empty
If first work item of a project is empty and the second one has a description, the copied notes will look like "; <description work2>". The prepended ; should not exists. ; should only be added between notes with content

DavidDamke:
fixes #146

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/ProjectReport.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ProjectReport.java
@@ -28,7 +28,7 @@ public class ProjectReport {
    /** The slf4j-logger for this class. */
    private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-   private int numberOfNotes;
+   private int numberOfNotEmptyNotes;
 
    private final int size;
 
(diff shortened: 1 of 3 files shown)


src/main/java/de/doubleslash/keeptime/view/ProjectReport.java
```diff
@@ -41,11 +41,11 @@ public ProjectReport(final int size) {
 
    public void appendToWorkNotes(final String currentWorkNote) {
       this.numberOfNotes++;
+      this.sb.append(currentWorkNote.trim());
       if (!currentWorkNote.isEmpty()) {
-         if (this.numberOfNotes > 1) {
+         if (this.size>1) {
```
Death111, line 46: we have tests which needs to be
a) checked if still working (pipeline fails, but not sure why. logs do not load currently)
b) extended with new behavior

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ProjectReport.java
```diff
@@ -41,14 +41,22 @@ public ProjectReport(final int size) {
 
    public void appendToWorkNotes(final String currentWorkNote) {
       this.numberOfNotes++;
+      this.sb.append(currentWorkNote.trim());
```
Death111, line 44: not sure why you append something and one line later we check if it is empty? so we do add also empty strings?

src/main/java/de/doubleslash/keeptime/view/ProjectReport.java
```diff
@@ -41,14 +41,22 @@ public ProjectReport(final int size) {
...
             this.sb.append(NOTE_DELIMETER);
          }
-         this.sb.append(currentWorkNote.trim());
       } else {
          LOG.debug("Skipping empty note.");
       }
+
+      if(numberOfNotes==this.size){
+         int lastSemiColonIndex = this.sb.lastIndexOf(";");
+         if (lastSemiColonIndex != -1) {
```
Death111, line 55: this logic seems really complex. I am also wondering how this works with the "lastIndexOf" and "deleteCharAt" "setLength" methods. quite confusing (for not adding a ; at the start).

As we have nice tests already I suggest to redo the whole logic here. Using class StringJoiner instead of StringBuilder ;) (I found that by accident. I know there exists some helper function to join a list of strings with a specified delimiter - but found this class which does the same thing basically.

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ProjectReport.java
DavidDamke: I changed the logic it is much simpler now and works.

DavidDamke: Can I delete this getter ?    
public int getNumberOfNotes() {
      return this.numberOfNotes;
I it is not used and we do not include the numberOfNotes anymore

Death111: if not needed we can remove it :)

Death111 (approved):

PR #149, opened 2023-01-23
focus on right textfields

### LINKED ISSUE #148: Validate that all dialogs have a default input field which has the focus
e.g. in "add project" dialog the "project name" should be focused as default. Currently, it has no field selected and one needs to click into it at first.
Check for all such dialogs.

DavidDamke:
- Edit project has focus on Description
- Add project has focus on Name

fixes #148

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/ManageProjectController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ManageProjectController.java
@@ -16,6 +16,7 @@
 
 package de.doubleslash.keeptime.view;
 
+import javafx.application.Platform;
 import javafx.beans.binding.Bindings;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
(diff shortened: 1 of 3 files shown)


src/main/java/de/doubleslash/keeptime/view/ManageProjectController.java
```diff
@@ -73,6 +74,14 @@ private void initialize() {
       sortIndexSpinner.getValueFactory().setValue(model.getAvailableProjects().size());
       formValidProperty.bind(Bindings.createBooleanBinding(() -> !nameTextField.getText().isBlank(),nameTextField.textProperty()));
       validateTextAlert.visibleProperty().bind(formValidProperty.not());
+      
+      Platform.runLater(() ->{
```
Death111, line 78: let's also have a look on Tab-order of the controls. If cursor is in "project name" and I press tab, the cursor will be on "color" instead of the next field "description".
also in edit work in report dialog the tab-order is not right

Death111 (changes requested):

Death111 (approved):

Issue #152, opened 2023-02-09
No version nr. in name of jar file


swa-ds:
Please leave out the version of KeepTime in the jar file and keeptime.bat. This will make updates much easier (just replace the jar file).

PR #153, opened 2023-02-10
#152: removed version number from .jar included in zip file. added ke…

Death111:
…eptime.bat to zip file.


Death111:
I accidentally deleted the version parameter in #129 - currently the version i not changed. no need to review this PR yet

swa-ds (approved):

PR #166, opened 2023-09-18
Feature/copy project name in report

### LINKED ISSUE #158: Add possibility to copy project name in Report
In the report view we should have the possibility to also copy the project name to clipboard. Not only the notes.

DavidDamke:
#158


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -263,7 +263,8 @@ private void updateReport(final LocalDate dateToShow) {
          }
 
          final HBox projectButtonBox = new HBox();
-         projectButtonBox.getChildren().add(createCopyProjectButton(onlyCurrentProjectWork));
+         projectButtonBox.getChildren().add(createCopyProjectNotesButton(onlyCurrentProjectWork));
+         projectButtonBox.getChildren().add(createCopyProjectNameButton(onlyCurrentProjectWork));
```
Death111: why pass work items from the project instead of the project name directly?

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -436,8 +439,29 @@ private Button createCopyProjectButton(final List<Work> projectWork) {
...
+   private Button createCopyProjectNameButton(final List<Work> projectWork) {
+      final Button copyProjectNameButton = new Button("", SvgNodeProvider.getSvgNodeWithScale(RESOURCE.COPY_PROJECT_NAME, 0.03, 0.03));
+      copyProjectNameButton.setMaxSize(20, 18);
+      copyProjectNameButton.setMinSize(20, 18);
+      copyProjectNameButton.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+      copyProjectNameButton.setTooltip(new Tooltip("Copy Project Name"));
+
+      final EventHandler<ActionEvent> eventListener = actionEvent -> {
+         LOG.debug("Copy to Clipboard clicked.");
+         final ProjectReport pr = new ProjectReport();
```
Death111: don't think we need a project report for a single string

src/main/java/de/doubleslash/keeptime/common/Resources.java
```diff
@@ -67,7 +67,8 @@ public enum RESOURCE {
 
       SVG_LICENSES_ICON("/svgs/closed-captioning.svg"),
 
-      ICON_MAIN("/icons/icon.png")
+      ICON_MAIN("/icons/icon.png"),
+      COPY_PROJECT_NAME("/svgs/copy.svg")
```
Death111: - inconsistent naming to other .svgs
- inconsitent formatting compared to other constants
- also please put the new svg to the other svgs

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: changed it so, now only the name gets passed

src/main/java/de/doubleslash/keeptime/view/ReportController.java
DavidDamke: removed it

src/main/java/de/doubleslash/keeptime/common/Resources.java
DavidDamke: changed it to match the other SVGs

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -263,7 +263,8 @@ private void updateReport(final LocalDate dateToShow) {
          }
 
          final HBox projectButtonBox = new HBox();
-         projectButtonBox.getChildren().add(createCopyProjectButton(onlyCurrentProjectWork));
+         projectButtonBox.getChildren().add(createCopyProjectNotesButton(onlyCurrentProjectWork));
+         projectButtonBox.getChildren().add(createCopyProjectNameButton(onlyCurrentProjectWork.get(0).getProject().getName()));
```
Death111: I'd say using the project name directly is easier and also more readable. What do you think?
```suggestion
         projectButtonBox.getChildren().add(createCopyProjectNameButton(project.getName()));
```

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -416,11 +417,13 @@ private GridPane setUpEditWorkGridPane(final Work work, final Dialog<Work> dialo
       return grid;
    }
 
-   private Button createCopyProjectButton(final List<Work> projectWork) {
-      final Button copyButton = new Button("", SvgNodeProvider.getSvgNodeWithScale(RESOURCE.SVG_CLIPBOARD, 0.03, 0.03));
-      copyButton.setMaxSize(20, 18);
-      copyButton.setMinSize(20, 18);
-      copyButton.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+   private Button createCopyProjectNotesButton(final List<Work> projectWork) {
+      final Button copyNotesButton = new Button("", SvgNodeProvider.getSvgNodeWithScale(RESOURCE.SVG_CLIPBOARD, 0.03, 0.03));
```
Death111: When using it the new icon feels more like "copy all child items" for me, as it has multiple instances

image

what do you think? I think I would exchange the icons. So the "clipboard" copies the current row (like it also does for the single work notes) and the "multiple clipboards" copies the children.

This also highlights that the new SVG Ressource name does not specify what the SVG itself shows, but the intended use. Which is not so nice as now we cannot switch the icons easily. We will also need to rename the icon resource.

DavidDamke: Switched the functionality.
Changed the Svg name to match the displayed svg

Death111 (approved):

PR #168, opened 2023-09-22
upgrade Spring Boot to 3.1.3

Death111:
Update to Spring > 3.0 as 3.0 is eol in November https://endoflife.date/spring-boot

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/SettingsController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/SettingsController.java
@@ -391,10 +391,13 @@ public class SettingsController {
             final String url = applicationProperties.getSpringDataSourceUrl();
             final String username = applicationProperties.getSpringDataSourceUserName();
             final String password = applicationProperties.getSpringDataSourcePassword();
-            // TODO: add an option at the next release to make the "FROM_1X flag" configurable. E.g. if we upgrade (in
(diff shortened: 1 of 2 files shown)


src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -391,10 +391,13 @@ private void initImportButton() {
             final String url = applicationProperties.getSpringDataSourceUrl();
             final String username = applicationProperties.getSpringDataSourceUserName();
             final String password = applicationProperties.getSpringDataSourcePassword();
-            // TODO: add an option at the next release to make the "FROM_1X flag" configurable. E.g. if we upgrade (in
-            // the release after) from H2 version 2.x to 2.x we must not set the "FROM_1X flag".
-            new RunScript().runTool("-url", url, "-user", username, "-password", password, "-script", file.toString(),
-                  "-options", "FROM_1X");
+
+            if (file.getName().contains("version-1")) {
+               new RunScript().runTool("-url", url, "-user", username, "-password", password, "-script", file.toString(),
```
Death111, line 396: When using filename (which the user may change while exporting, but is good enough for me) I would check for contains "H2-version-1." (to make the check more obvious what version it checks).
Also in the if branch please add a info log statement to log out that the FROM_1X feature is used.

DavidDamke: Add a log info statement

PR #179, opened 2025-01-25
Feature/#178 heimat integration

### LINKED ISSUE #178: Add integration with "Heimat"
- Allow to fetch projects from Heimat
- Allow to save times to Heimat

Death111:


src/main/java/de/doubleslash/keeptime/model/ExternalProjectMapping.java
```diff
@@ -0,0 +1,111 @@
+// Copyright 2019 doubleSlash Net Business GmbH
```
Death111, line 1: update header and add it also to new classes

src/main/java/de/doubleslash/keeptime/rest/integration/heimat/HeimatAPI.java
```diff
@@ -0,0 +1,99 @@
...
+                                                              .body(heimatTime)
+                                                              .retrieve()
+                                                              .onStatus(HttpStatus.UNAUTHORIZED::equals,
+                                                                    (request, response) -> {
+                                                                       throw new UnauthorizedException();
+                                                                    })
+                                                              .toEntity(String.class);
+
+      System.out.println("Status Code: " + responseEntity.getStatusCode());
+      System.out.println("Response Body: " + responseEntity.getBody());
```
Death111: remove sysouts

Death111:
Merging besides having a failing Sonar. Sonar complains about test-coverage which is fair. But for the UI-Controllers we do have a test gap for every controller. For non-ui logic we do have tests in place.

PR #183, opened 2025-05-14
#178: Fix NPE when Heimat Task is no longer available for a day

Death111:
* fix issue when task is no longer available for a present time. showing id instead of friendly project name instead.


sesturm:
bugbot run

PR #184, opened 2025-09-05
PTBAS-737: using consistent english language now

MomdAli:
**Locale management and formatting improvements:**

* Added `setSystemLocale`, `getSystemLocale`, and `applySystemLocaleOnDate` methods to the `DateFormatter` utility class, allowing centralized management of the application's locale and consistent formatting of dates in `DatePicker` components. (`DateFormatter.java`) [[1]](diffhunk://#diff-1ac6d1ddce45b73b20d37f63df7255a262a07f302ee53ae8d976cf42883978abR19-R35) [[2]](diffhunk://#diff-1ac6d1ddce45b73b20d37f63df7255a262a07f302ee53ae8d976cf42883978abR66-R94)
* On application startup, the system locale is now set in `DateFormatter`, and the default locale is explicitly set to English to ensure predictable behavior. (`App.java`) [[1]](diffhunk://#diff-3439823e632a1b7d0d15b927980d2db5b02f094cedbf4ab438615f761a83d84cR24-R28) [[2]](diffhunk://#diff-3439823e632a1b7d0d15b927980d2db5b02f094cedbf4ab438615f761a83d84cR89-R90)

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/App.java
+++ b/src/main/java/de/doubleslash/keeptime/App.java
@@ -21,9 +21,11 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.time.LocalDate;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import de.doubleslash.keeptime.common.DateFormatter;
 import javafx.stage.Window;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -84,6 +86,8 @@ public class App extends Application {
    @Override
    public void init() throws Exception {
       LOG.info("Starting KeepTime.");
+      DateFormatter.setSystemLocale(Locale.getDefault());
+      Locale.setDefault(Locale.ENGLISH);
       final DefaultExceptionHandler defaultExceptionHandler = new DefaultExceptionHandler();
       defaultExceptionHandler.register();
 
--- a/src/main/java/de/doubleslash/keeptime/common/DateFormatter.java
+++ b/src/main/java/de/doubleslash/keeptime/common/DateFormatter.java
@@ -16,14 +16,22 @@
 
 package de.doubleslash.keeptime.common;
 
+import javafx.scene.control.DatePicker;
+import javafx.util.StringConverter;
+
 import java.time.Duration;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.format.FormatStyle;
+import java.util.Locale;
 
 public class DateFormatter {
    private static DateTimeFormatter dayDateFormatter = DateTimeFormatter.ofPattern("eeee dd.MM.yyyy");
    private static DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss");
+
+   private static Locale systemLocale = Locale.GERMAN;
 
    private DateFormatter() {
       throw new IllegalStateException("Utility class: DateFormatter");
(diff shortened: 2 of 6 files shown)


MomdAli:
@CodiumAI-Agent /review

src/main/java/de/doubleslash/keeptime/common/DateFormatter.java
```diff
@@ -16,15 +16,23 @@
...
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.format.FormatStyle;
+import java.util.Locale;
 
 public class DateFormatter {
    private static DateTimeFormatter dayDateFormatter = DateTimeFormatter.ofPattern("eeee dd.MM.yyyy");
    private static DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss");
 
+   private static Locale systemLocale = Locale.GERMAN;
```
Death111: could we initialize this with `Locale.getDefault()`? we would not need the setter in this case
```suggestion
   private static Locale systemLocale = Locale.getDefault();
```

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/App.java
```diff
@@ -21,9 +21,11 @@
 import java.io.StringWriter;
 import java.time.LocalDate;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.stream.Collectors;
```
Death111, line 27: Sync dialog shows AM:
grafik

Death111 (changes requested):

src/main/java/de/doubleslash/keeptime/common/DateFormatter.java
MomdAli: This doesn't work because the static variable is initialized when the class is loaded, so it won't reflect locale changes made later in init().

src/test/java/de/doubleslash/keeptime/common/DateFormatterTest.java
```diff
@@ -38,4 +43,21 @@ void zeroSecondsBetweenTest() {
...
+      DatePicker datePicker = new DatePicker();
+      LocalDate date = LocalDate.of(2024, 6, 3);
+
+      // WHEN
+      DateFormatter.applySystemLocaleOnDate(datePicker);
+      datePicker.setValue(date);
+
+      // THEN
+      String shown = datePicker.getEditor().getText();
+      assertTrue(shown.contains("Montag") || shown.contains("06.2024") || shown.contains("06"));
```
Death111: assert is weird with 3 items in OR condition -> only one must be true?
Better create three asserts with each aspect.

src/test/java/de/doubleslash/keeptime/common/DateFormatterTest.java
```diff
@@ -38,4 +43,21 @@ void zeroSecondsBetweenTest() {
       final long secondsBewtweenSwitched = DateFormatter.getSecondsBewtween(endDate, startDate);
       assertThat(secondsBewtweenSwitched, Matchers.is(0l)); // why??
    }
+
+   @Test
+   void applySystemLocaleShouldSetDatePickerLocale() {
+      // GIVEN
+      Locale.setDefault(Locale.ENGLISH);
```
Death111: this is "dangerous" as you do not reset the locale back after the test. So the next test may fail because the locale is still english.
should make sure to get the system into previous state if modifying something which may influence other tests.

src/main/java/de/doubleslash/keeptime/common/DateFormatter.java
Death111: Sounds like it behaves exactly like I wanted it to behave? Or what am I missing? I want it to have the system default language set (not reflecting locale changes made later in init)?

Death111 (changes requested):

Death111:
@MomdAli I was using it the last weeks and I did not notice issues - only annoying thing is the calendar view that sunday is the first day of the week. can this be changed / also localized somehow?
grafik

src/main/java/de/doubleslash/keeptime/common/DateFormatter.java
```diff
@@ -55,4 +63,36 @@ public static String toTimeString(final LocalDateTime localDateTime) {
...
+               return (s != null && !s.isEmpty()) ? LocalDate.parse(s, formatter) : null;
+            } catch (DateTimeParseException e) {
+               return null;
+            }
+         }
+      });
+
+      datePicker.setOnShowing(e-> Locale.setDefault(Locale.Category.FORMAT, systemLocale));
+      datePicker.setOnHiding(e-> Locale.setDefault(Locale.Category.FORMAT, Locale.getDefault()));
+      datePicker.setOnAction(e-> Locale.setDefault(Locale.Category.FORMAT, Locale.getDefault()));
```
Death111, line 95: these resets did not make sense for me, as you reset it to the "wrong" locale. But it showed me this way to only set the locale for FORMAT. I created #191 to most likely replace this PR - lets review it.

Death111:
Implemented via #191

PR #186, opened 2025-09-12
PTBAS-741: Adjust Mapping Dialog

MomdAli:
# mapping dialog
* user can now choose to remove invalid mappings, or keep them
* invalid mappings are also highlighted in ui
* added filter possibility
# sync dialog
* fixed issue which closed sync dialog 'directly' after it was openend
* fixed issue when same project was shown twice in the sync dialog
# setting dialog
* highlights heimat token expiration in red when expired


src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
```diff
@@ -169,22 +170,66 @@ public List<Mapping> getTableRows(final LocalDate currentReportDate, final List<
 
          if (optionalExistingMapping.isPresent()) {
             final Mapping existingMapping = optionalExistingMapping.get();
+
+            // Ensure we merge projects robustly: include any projects mapped to the same Heimat task
```
Death111: I think I would like to only have projects shown in the dialog, which I also worked on at that date. not all projects which are mapped to it

Death111: hm okay. but when I dont have the time in keeptime but heimat, what should it show for keeptime projects? not so sure if no projects would be wanted (which would be the case when adapting as I wanted it)

src/test/java/de/doubleslash/keeptime/controller/HeimatControllerTest.java
```diff
@@ -516,4 +519,41 @@ void shouldReturnErrorsWhenErrorsOccurred() {
    }
    // shouldOnlyUpdateHeimatWhenSomethingHasChanged (not needed - user should decide)
 
+   @Test
+   void shouldNotCreateDuplicateHeimatEntryWhenMultipleProjectsMappedAndSomeHaveWork() {
```
Death111: I do not think that this "GIVEN" created the issue observed. Running this test on previous code fails yes - but not because the report has 2 rows in the end - which was the issue that we see the same twice.

It failse because the other project is not contained in the projects (Matchers.containsInAnyOrder(workProject1, workProject2)).

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
```diff
@@ -469,5 +537,33 @@ public HeimatTask getHeimatTask() {
...
+      public boolean isPendingRemoval() {
+         return pendingRemoval;
+      }
+
+      public void setPendingRemoval(boolean pendingRemoval) {
+         this.pendingRemoval = pendingRemoval;
+      }
+   }
+
+   public List<HeimatTask> getAllKnownHeimatTasks(final LocalDate forDate) {
```
Death111: I'm not sure why we need this. It is referenced in some places - but not sure why we need it. need to dig deeper to understand why we need this

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -215,6 +216,9 @@ public class SettingsController {
    @FXML
    private Label heimatExpiresLabel;
 
+   @FXML
+   private Label expirationDateLabel;
```
Death111, line 217: we can write all info into existing heimatExpiresLabel? or why do we need a 2nd label for this?

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -411,9 +415,21 @@ private void initializeHeimat() {
...
+               heimatExpiresLabel.setTextFill(Color.BLACK);
+               expirationDateLabel.setTextFill(Color.BLACK);
+            }
+
+            expirationDateLabel.setText(jwt.expiration().toString());
+
          } catch(Exception e){
-            heimatExpiresLabel.setText("Does not seem to be valid");
+            heimatExpiresLabel.setText("");
+            expirationDateLabel.setText("Does not seem to be valid");
```
Death111, line 427: could also set it to color red if its not valid

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
Death111: ah I see. the label holds the row-name for the setting **"Expires:"** \<value\> currently. Still, I would keep this as is and ad the Expired info to the value displayed.
Feels a little uncommon that the value description changes for me :)

src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
```diff
@@ -86,15 +89,29 @@ private void initialize() {
...
       final ExistingAndInvalidMappings existingAndInvalidMappings = heimatController.getExistingProjectMappings(
             externalProjects);
-      final List<HeimatController.ProjectMapping> previousProjectMappings = existingAndInvalidMappings.validMappings();
 
+
+      final List<HeimatController.ProjectMapping> previousProjectMappings = existingAndInvalidMappings.validMappings();
       final ObservableList<HeimatController.ProjectMapping> newProjectMappings = FXCollections.observableArrayList(
             previousProjectMappings);
+
+      Platform.runLater(() -> {
```
Death111, line 113: runLater with a dialog which modifies the ui 'indirectly' seems not soo nice. not sure if we could solve this differently

Death111 (changes requested):

Death111:
in sync dialog times which exist in heimat but not in keeptime the projects are not shown in bold. might fix in this pr

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
MomdAli: The method gets tasks from the API for that date (via getTasks), and also includes tasks from stored mappings (even if not available today) and merges both.

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
MomdAli: Done

src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
MomdAli: I changed it to use on thisStage.setOnShown, however I don"t know if that is a good idea, so I will leave it but if you want to revert it then feel free to tell me.

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
MomdAli: Done

src/test/java/de/doubleslash/keeptime/controller/HeimatControllerTest.java
MomdAli: Done

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
MomdAli: Done

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
Death111: should also remove the label reference and not write a static value "Expires:" to it..

Death111: removed it

src/test/java/de/doubleslash/keeptime/controller/HeimatControllerTest.java
Death111: the newly created test `shouldNotCreateDuplicateHeimatEntryWhenMultipleProjectsMappedAndBothHaveWork` is also already green in develop state. so it does not reproduce the issue we want to fix.

it also duplicates this test `shouldCombineAllWorksAndProjectsWhenMultipleProjectsAreMappedToSame`

Death111: I removed the duplicated test, added new test which fails in develop but works in this branch (reproducing the issue we wanted to fix)

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
```diff
@@ -26,15 +26,14 @@
 import de.doubleslash.keeptime.rest.integration.heimat.model.HeimatTask;
```
Death111, line 37: I guess easiest to revert this file.
looks quite complicated to mostly fix the bug with duplicate rows (which should be only needed in one place.

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
```diff
@@ -391,50 +427,53 @@ private String getAsJson(final HeimatTask heimatTask) {
...
          }
+         final ExternalProjectMapping mapped = mapping.get();
+
+         HeimatTask mappedTask = getHeimatTaskFromMapping(mapped);
+
          final Optional<HeimatTask> any = externalProjects.stream()
-                                                          .filter(ep -> ep.id() == mapping.get().getExternalTaskId())
+                                                          .filter(ep -> ep.id() == mapped.getExternalTaskId())
                                                           .findAny();
          if (any.isEmpty()) {
```
Death111, line 412: as this method is called with externalProjects = heimatController.getAllKnownHeimatTasks()
which gets all from heimat+persisted ones, 

it will never be empty? so it will never remove anything I assume

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
Death111: done

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
Death111: I fixed the caller to only pass actually available tasks

src/main/java/de/doubleslash/keeptime/controller/HeimatController.java
Death111: not referenced anymore - deleted it (previous logic did not make sense for invalid mapping detection, where existing+non existing tasks were passed from this method)

src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
Death111: I reverted it to platform.runlater. both are not so nice..

src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
```diff
@@ -103,58 +130,61 @@ private void initialize() {
...
+               Tooltip tooltip = new Tooltip(item);
+               setTooltip(tooltip);
+            }
+         }
+      });
+
       // External Project column with dropdown
       final ObservableList<HeimatTask> externalProjectsObservableList = FXCollections.observableArrayList(
             externalProjects);
-      externalProjectsObservableList.add(0, null); // option to clear selection
```
Death111, line 109: todo check if we can readd this for easier selection

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -411,9 +412,20 @@ private void initializeHeimat() {
       heimatPatTextField.textProperty().addListener((observable, oldValue, newValue)->{
          try{
             final JwtDecoder.JWTTokenAttributes jwt = JwtDecoder.parse(newValue);
-            heimatExpiresLabel.setText(jwt.expiration().toString());
+            final boolean isExpired = !JwtDecoder.isExpired(jwt, LocalDateTime.now());
```
Death111: isExpired = !JwtDecoder.isExpired

looks weird. why invert isExpired and assign it isExpired

src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
Death111: not so easy...

but with current implementation one cannot remove a mapping (set no mapping). Need to add possibility to select <none>

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
Death111: fixed it

Death111 (changes requested):

src/test/java/de/doubleslash/keeptime/controller/HeimatControllerTest.java
```diff
@@ -516,4 +526,64 @@ void shouldReturnErrorsWhenErrorsOccurred() {
...
+      final HeimatTime existingTime = new HeimatTime(project2To1Mapping.getExternalTaskId(), now.toLocalDate(), null,
+              null, 15, "Notes 2", 99);
+      when(mockedHeimatAPI.getMyTimes(now.toLocalDate())).thenReturn(Arrays.asList(existingTime));
+
+      // ACT
+      final List<HeimatController.Mapping> tableRows = heimatController.getTableRows(now.toLocalDate(), workItems);
+
+      // ASSERT
+      assertThat("There should be only one row for this task.",tableRows.size(), Matchers.is(1));
+   }
```
Death111, line 713: check if we can do a test with the new keepInvalidMappings feature

Death111: created

Death111 (approved):

src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
Death111: added support for non selection

Issue #188, opened 2025-09-29
Enhancement: Add ability to manually create tasks and schedule them


MomdAli:
While editing work entries manually is already supported, there is currently no way to create new work entries from scratch without project switching. This enhancement should provide a user-friendly interface for manually adding work entries directly, supporting both immediate and scheduled (future) entries.

### Acceptance Criteria
- Users can manually create new work entries without switching projects.
- Users may optionally schedule the work entry for a future time.

PR #190, opened 2025-10-09
#188: Add Ability to Manually Add Work Entries

MomdAli:
Add the ability to add an "Add Work" button next to the Sync button in the report view. When the user clicks Add Work the Manage Work popup appears, allowing them to enter details for a new work entry. If the user confirms by pressing OK, the new work is saved and immediately shown in the report.

- The button uses the same Manage Work dialog as editing, so the user experience is consistent.
- The logic for saving is simple and reuses existing methods.
- Changed the style of the time picker in mange work menu for easier use.

### DIFF
--- a/src/main/java/de/doubleslash/keeptime/view/ReportController.java
+++ b/src/main/java/de/doubleslash/keeptime/view/ReportController.java
@@ -18,6 +18,8 @@ package de.doubleslash.keeptime.view;
 
 import java.io.IOException;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.util.*;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


src/main/java/de/doubleslash/keeptime/view/ReportController.java
```diff
@@ -132,10 +137,34 @@ private void initialize() {
       colorTimeLine = new ColorTimeLine(colorTimeLineCanvas);
 
       expandCollapseButton.setOnMouseClicked(event ->toggleCollapseExpandReport());
+      if (addWorkButton != null) {
```
Death111, line 140: check not needed

src/main/java/de/doubleslash/keeptime/view/TimePickerComboBox.java
```diff
@@ -0,0 +1,116 @@
+package de.doubleslash.keeptime.view;
```
Death111, line 1: I think I wanted to revert the timepicker changes. only keep the add work button.

Death111 (commented):
I created !201 which only contains the button. I am not sure about the new datepicker elements

PR #191, opened 2025-10-20
PTBAS-737: using consistent english language now - strategy 2

Death111:
On application startup, the system Locale is now set to ENGLISH if language is not already ENGLISH. However, the Locale used for formatting e.g. dates is kept in Systems default Locale.


Death111:
@MomdAli can you please review this PR? I think it works like we want it to much simpler than #184

MomdAli (approved):
I have looked into it and it should be working as expected.

PR #199, opened 2025-11-26
Feature/#198 password encoding

mplieske:
I added Argon2 password encoding.

Take it or leave it :laughing: 

Any suggestions are welcome :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -168,6 +168,12 @@
       <version>2.18.2</version>
     </dependency>
 
+    <!-- for Argon2PasswordEncoder -->
+    <dependency>
+        <groupId>org.bouncycastle</groupId>
+        <artifactId>bcprov-jdk18on</artifactId>
+        <version>1.78.1</version> <!-- or latest -->
+    </dependency>
   </dependencies>
   <build>
     <finalName>keeptime-${project.version}</finalName>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -168,6 +168,12 @@
       <version>2.18.2</version>
     </dependency>
 
+    <!-- for Argon2PasswordEncoder -->
+    <dependency>
+        <groupId>org.bouncycastle</groupId>
+        <artifactId>bcprov-jdk18on</artifactId>
```
Death111, line 174: I guess I would rather use a spring buildin passwordencoder like https://docs.spring.io/spring-security/reference/features/authentication/password-storage.html (I did not validate this yet.) But I like less dependencies :P

src/main/java/de/doubleslash/keeptime/view/SettingsController.java
```diff
@@ -758,12 +759,15 @@ private void handleApiOn() {
       String username = authName.getText();
       String password = authPassword.getText();
```
Death111, line 760: like you commented in the issue
"The configuration screen reads the password from the application.properties and loads it into the view. If you already have a configuration for the REST API feature and you want to change something, eg the port, the already hashed password will be hashed and saved."

Your idea -> "Probably we should only save a password if provided a new one and we should not load the password from the application.properties into the view"

sounds reasonable, yes 👍

Death111 (changes requested):

pom.xml
mplieske: The Argon2PasswordEncoder is mentioned on the page you linked :)
https://docs.spring.io/spring-security/reference/features/authentication/password-storage.html#authentication-password-storage-argon2

To be fair, we could use the default password encoder. Probably it will be sufficient for KeepTime 😄

Death111: you are right. argon is mentioned

the default one sound also sufficient - saves 8MB of disk space :D

PR #206, opened 2026-07-08
Heimat-Support: Added a tooltip containing the task name.

w-lutz:


src/main/java/de/doubleslash/keeptime/view/ExternalProjectsMapController.java
```diff
@@ -253,6 +253,8 @@ protected void updateItem(HeimatTask item, boolean empty) {
                setText(null);
             } else {
                setText(item.taskHolderName() + " - " + item.name());
```
Death111, line 255: we might also want to allow resizing of the popup, as hovering over each project is also not too great

